PR #41, opened 2016-01-06
Fix #40: Remove panes and set up: Tab = next unread thread

### LINKED ISSUE #40: (feature) go to next unread thread in thread index
(Sorry for launching so many issues and feature requests at you, I'm posting them as they go when I use astroid and I just mean to record them)

In the thread index view, I don't see any keybinding that will take me to the next unread thread. I think that would be useful.

For instance in mutt's mail folder view (which is equivalent to an index view in astroid for the user), Tab will automatically highlight the next unread email. So Tab+Enter will usually be a quick combination to read all unread emails in a folder.

gauteh:
#40: Adds the keybinding Tab to go to the next _unread_ thread.

This overwrites the pane swap. I am considering to remove the pane stuff, its function can be accomplished by opening a new window. Comments?

### DIFF
--- a/src/modes/thread_index/thread_index.cc
+++ b/src/modes/thread_index/thread_index.cc
@@ -68,10 +68,9 @@ namespace Astroid {
           return true;
         });
 
-    /*
-    keys.register_key (Key ((guint) GDK_KEY_Tab), "thread_index.pane_swap_focus",
+    keys.register_key (Key (false, true, (guint) GDK_KEY_Tab), "thread_index.pane_swap_focus",
         "Swap focus to other pane if open",
-        [&] (Key k) {
+        [&] (Key) {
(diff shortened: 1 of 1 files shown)


apm256:
I do not use the pane feature, and I agree that opening a new window seems enough. So, no problem for me.

gauteh:
@hugoroy writes (#40):
> I thought panes were nice. Although I don't use them for now, I think it would be sad to drop panes (especially for the future, when porting to systems where the windows manager sucks).

There has been a bug for a while that prevents focus to shift back to
the thread-index, so I don't think many users use them. Anyway, I like
it better with the Tab key bound to next unread. Can find another key
for the panes and keep them a little longer at least.

gauteh:
Merged, panes prevail.

Issue #29, opened 2016-01-03
Hooks and user's own macros?


hugoroy:
Related to issue #21 but this isn't stricly about keybindings but more about user-defined macros:

For instance, I'd like to reasign `a` to not only toggle the "inbox" tag but also to toggle an "archive" tag.

In mutt (the kz fork with notmuch support), the keybinding is actually a macro:
`macro index,pager   a       "<modify-labels-then-hide>-unread -inbox +archive<enter><sync-mailbox>"             "archive"`

Is this something that can already be done by the user like keybindings?

gauteh:
There was a brief discussion about this in #25, on setting up
user-defined hooks.

It is currently not possible, but I was planning to allow Python hooks
to be configured. Or possibly just regular shell scripts. The script
could get the message id or thread id and do something with it, return
true or false and astroid will re-render the line depending on that.

In shell; just use regular notmuch commands to do whatever.
In python; use the notmuch python bindings - or just execute a notmuch
command.

Shell is way easier to implement.

This would have to be hooked into the keybindings somehow, I was
thinking maybe:

thread_index.hook(my-special-hook)=a

along with some pre-defined hooks that can be attached to other
(non-key) events?

I currently don't really have a command/macro system.

What do you think?

hugoroy:
Well, I wouldn't need anything fancy and so using shell for notmuch commands would make me very  happy :-) / for instance:

`thread_index.hook(notmuch tag -inbox -unread +archive)=a`

gauteh:
Hugo Roy writes on January 3, 2016 21:36:
> Well, I wouldn't need anything fancy and so using shell for notmuch commands would make me very  happy :-) / for instance:
> 
> `thread_index.hook(notmuch tag -inbox -unread +archive)=a`

Yeah, this seems like a good idea, but perhaps use something like:

  main_window.run
  thread_index.run
  thread_index.multi.run

with pre-defined arguments.

hugoroy:
I don't know. That looks like another configuration file needed to pre-define the arguments. I think it's alright if it's not pre-defined. But I don't have strong arguments any way.

gauteh:
What I mean is that you can configure e.g.:

`thread_index.run(notmuch tag -inbox -unread +archive)=a`

which will run the command with one argument which is the thread id of
the currently selected thread:

`notmuch tag -inbox -unread +archive thread:00000xxx`

but you could also use your own script:

[...]

PR #46, opened 2016-01-08
toggle waiting tag

hugoroy:
Tagging emails where you are waiting for an answer is useful.

This adds the "w" keybinding for toggling the "waiting" tag.

### DIFF
--- a/src/modes/thread_index/thread_index_list_view.cc
+++ b/src/modes/thread_index/thread_index_list_view.cc
@@ -646,6 +646,18 @@ namespace Astroid {
           return true;
         });
 
+    keys->register_key ("w", "thread_index.waiting",
+        "Toggle 'waiting' tag on thread",
+        [&] (Key) {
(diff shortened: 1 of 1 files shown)


gauteh:
Ideally this should be done through #29 I think. Since the "waiting" tag
is a user-specific way of using notmuch I am a bit reluctant to merging
it (I for instance use a "queue" tag), otherwise the code looks just fine.

- gaute

hugoroy:
↪ janvier 9, 2016 5:16 , Gaute Hope:
> Ideally this should be done through #29 I think. Since the "waiting" tag
> is a user-specific way of using notmuch I am a bit reluctant to merging
> it (I for instance use a "queue" tag), otherwise the code looks just fine.

Yes, this makes sense :)

gauteh:
Check out #76 :)

hugoroy:
Awesome :-)

PR #47, opened 2016-01-09
Allow sent tags to be configured

hugoroy:
I prefer to add an inbox tag to messages I send. That way I can see when I haven't got an answer, etc. This adds an inbox tag to the sent tag of sent emails. Ideally of course, this shouldn't be hardcoded but, like "excluded_tags" in the config file.

### DIFF
--- a/src/db.hh
+++ b/src/db.hh
@@ -103,7 +103,8 @@ namespace Astroid {
       void load_tags ();
       void test_query ();
 
-      vector<ustring> sent_tags = { "sent" };
+      /* FIXME: add astroid.notmuch.sent_tags in config */
+      vector<ustring> sent_tags = { "sent,inbox" }; 
       vector<ustring> draft_tags = { "draft" };
       void add_sent_message (ustring);
       void add_draft_message (ustring);


gauteh:
Yeah, I agree. I don't think I can merge this before it is configurable,
as you mention this can be done exactly like excluded_tags. I think the
default in config.cc should be only the "sent" tag. It is not necessary
to do any configuration migration since adding a config option will be
automatically merged in.

- gaute

gauteh:
Gaute Hope writes on January 9, 2016 17:14:
> Yeah, I agree. I don't think I can merge this before it is configurable,
> as you mention this can be done exactly like excluded_tags. I think the
> default in config.cc should be only the "sent" tag. It is not necessary
> to do any configuration migration since adding a config option will be
> automatically merged in.

Perhaps it should be a per-account configuration option?

hugoroy:
↪ janvier 9, 2016 5:28 , Gaute Hope:
> Perhaps it should be a per-account configuration option?

I think most configuration options should be general, with per account
options allowing for overriding general configurations.

gauteh:
I've added it as a global and as an additional per account config option. Note that in your code you did not extend the vector, you just made changed the default tag into "sent,inbox" instead of "sent", "inbox". Please test #51.

hugoroy:
Ha, good to know :) I was far from sure this was going to work. I'll test later, gotta go.

hugoroy:
It works!

Issue #45, opened 2016-01-08
(feature) thread index order config: from oldest to newest


hugoroy:

gauteh:
Please test #50.

gauteh:
Fixed in #50.

PR #50, opened 2016-01-10
ti: C-s cycle through sort order for thread_index, default is configurable

gauteh:
#45

### DIFF
--- a/History.txt
+++ b/History.txt
@@ -22,6 +22,7 @@
 * Show name of pre-defined searches in tab.
 * Tab: jump to next unread thread.
 * X: Force close tab.
+* C-s: Cycle through sorting the thread_index (default configurable)
 
 == v0.4 / 2015-08-15
 
--- a/src/config.cc
+++ b/src/config.cc
@@ -117,6 +117,8 @@ namespace Astroid {
 
     /* ui behaviour */
     default_config.put ("thread_index.page_jump_rows", 6);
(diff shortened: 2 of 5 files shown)


hugoroy:
The keybinding works, however "oldest" does not seem to work as expected:
capture d ecran de 2016-01-10 12-58-58

gauteh:
It cycles through "oldest", "newest", "messageid", "unsorted". The default is newest so the one afterwards is messageid. Just check the log output.

gauteh:
(what font are you using btw?)

hugoroy:
Yes, I could see that. But when "oldest" is used, I can't really point out what is going on. The screenshot is an "oldest" sorting where a quite recent email was between two threads where there were no emails since yesterday.

(the font is http://input.fontbureau.com/ but it's not under a free license unfortunately -- just fine for personal use)

gauteh:
Are you configuring through the config file or using C-s?

gauteh:
I think this is because it is using the oldest message in the thread. Not the newest message, and then sort by thread. Not sure if it is possible to do it otherwise. It does look a bit weird in the thread-index. Could you check if there is an older message in the 5m thread?

hugoroy:
Yes, it appears that it's sorting by oldest thread. Actually, this is logical. I guess that what mutt is doing for instance is sorting by email and not by thread. 

I think it's good enough :-) there's no point in trying to replicate how it would work in email-based mua like mutt.

Issue #30, opened 2016-01-03
Crash when opening thread with mail containing forwarded email in attachment


hugoroy:
Steps to reproduce

1. Select an email with a forwarded attachment (message/rfc822)
2. Crash

```
[debug] (13:07:10): ky: ThreadIndex mode, handling: j (thread_index.next_thread)
[info ] (13:07:10): poll: periodic poll..
[debug] (13:07:10): poll: requested..
[info ] (13:07:10): poll: emitted poll state: ling: j (thread/home/hugo/.config/astroid/poll.sh1
[info ] (13:07:10): 
[info ] (13:07:10): db: open db read-only.
[debug] (13:07:10): db: open time: 0.869 ms.
[debug] (13:07:10): poll: revision before poll: 202279
[ERROR] (13:07:10): poll: poll script does not exist or is not a regular file.
[info ] (13:07:10): db: closing db.
[info ] (13:07:10): poll: emitted poll state: 0
[debug] (13:07:10): ky: Mode,  unknown key: Return
[debug] (13:07:10): ky: ThreadIndex mode, handling: Return (thread_index.open_thread)
[debug] (13:07:10): ti: open thread: 000000000001f15d (1)
[info ] (13:07:10): tv: using user html file: /home/hugo/.config/astroid/ui/thread-view.html
[debug] (13:07:10): tv: testing version: 1
[info ] (13:07:10): tv: using user css file: /home/hugo/.config/astroid/ui/thread-view.css
[debug] (13:07:10): tv: testing version: 0
[ERROR] (13:07:10): tv: css file version does not match!
[debug] (13:07:10): key: registering key: thread_view.down: j
[debug] (13:07:10): key: registering key: thread_view.next_element: C-j
[debug] (13:07:10): key: registering key: thread_view.scroll_down: J
[debug] (13:07:10): key: alias: Down(65364)
[debug] (13:07:10): key: registering key: thread_view.page_down: C-J
[debug] (13:07:10): key: alias: C-Down(65364)
[debug] (13:07:10): key: alias: 65366(65366)
[debug] (13:07:10): key: registering key: thread_view.up: k
[debug] (13:07:10): key: registering key: thread_view.previous_element: C-k
[debug] (13:07:10): key: registering key: thread_view.scroll_up: K
[debug] (13:07:10): key: alias: Up(65362)
[debug] (13:07:10): key: registering key: thread_view.page_up: C-K
[debug] (13:07:10): key: alias: C-Up(65362)
[...]

PR #55, opened 2016-01-13
mime-message: #30: add test case, not failing yet. failure happens in thread-view

gauteh:
test (and eventually fix) for #30.


gauteh:
Replaced by #56.

PR #57, opened 2016-01-13
Fix #15

### LINKED ISSUE #15: crash when using "tab" to complete a tag
Hi gauteh,

Here is a small bug:
- choose a thread with a sufficient number of tags;
- press 'l';
- type the beginning of an existing tag;
- complete it with the "TAB" key (nice feature by the way, thanks for it!);
- then, type "," to add another tag.

Result: astroid crashes with the following trace:

```
[debug] (16:21:37): cb: completion: taking next match
(astroid:21057): glibmm-ERROR **: 
what: basic_string::assign: __pos (which is 18446744073709551615) > this->size() (which is 24)
```

ff2000:
Do not try to substr the tail if we complete at the end of the
command bar. This will prevent overflows.

### DIFF
--- a/src/command_bar.cc
+++ b/src/command_bar.cc
@@ -374,8 +374,10 @@ namespace Astroid {
                                 // not the first
       newt += completion;
 
-      // add remainder of text field
-      newt += t.substr (pos+key.size() + ((pos > 0) ? 1 : 0), t.size());
+      // add remainder of text field only if we are not at the end of the line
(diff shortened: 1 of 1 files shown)


ff2000:
Feel free to change the commit message, I am not too good at that ;) (Don't even know if overflow is the right term...)

gauteh:
Excellent! Merged. (git commit --amend is your friend)

Issue #60, opened 2016-01-15
(AUR) failed to compile


hugoroy:
```
==> Lancement de check()...
scons: Reading SConscript files ...
building release: v0.4.r205.gcdea3bb
debug flag enabled: False
Checking for pkg-config... (cached) yes
Checking for gtkmm-3.0 >= 3.10... (cached) yes
Checking for glibmm-2.4... (cached) yes
Checking for gmime-2.6 >= 2.6.18... (cached) yes
Checking for webkitgtk-3.0... (cached) yes
Checking for libsass... (cached) yes
Checking for C library notmuch... (cached) yes
Checking for C function notmuch_database_get_revision()..(cached) yes
Checking for C++ library boost_filesystem... (cached) yes
Checking for C++ library boost_system... (cached) yes
Checking for C++ library boost_program_options... (cached) yes
writing src/build_config.hh..
scons: done reading SConscript files.
scons: Building targets ...
Running tests in test/test_theme
Running 1 test case...
[info ] (00:01:35): welcome to astroid! - v0.4.r205.gcdea3bb
[debug] (00:01:35): utf8: 1, UTF-8
[info ] (00:01:35): cf: loading test config.
[TEST ] (00:01:35): cf: using home directory: /tmp/yaourt-tmp-hugo/aur-astroid/src/astroid/test/test_home
[info ] (00:01:35): cf: test config, loading defaults.
[info ] (00:01:35): date: init.
[info ] (00:01:35): ac: initializing accounts..
[info ] (00:01:35): ac: setup account: charlie for Charlie Root (default: 1)
[info ] (00:01:35): global actions: set up.
[info ] (00:01:35): poll: setting up.
[debug] (00:01:35): poll: interval: 0
[info ] (00:01:35): poll: periodic polling disabled.
[debug] (00:01:35): theme: loading..
[debug] (00:01:35): tv: testing version: 2
[debug] (00:01:35): tv: testing version: 2
[info ] (00:01:35): theme: processing: ui/thread-view.scss
unknown location(0): fatal error: in "Theme/loading_theme": memory access violation at address: 0x023ec970: no mapping at fault address
test/test_theme.cc(16): last checkpoint: "loading_theme" entry.

*** 1 failure is detected in the test module "TestTheme"
scons: *** [test/test_theme.passed] Error 1
scons: building terminated because of errors.
```

ff2000:
Would like to reproduce, but I even don't get that far.
It already hangs at compiling the first test.
```
[...]

PR #61, opened 2016-01-16
remove "using namespace" from headers.

ff2000:
Also start cleaning up includes. The only "using" i left was for ptree in src/config.hh. I hope (and think) this won't do any damage. I could do something with that (namespace/type alias) if you want.

When there was an issue in a .cc-file I also tried to reduce pollution of global
namespace. Especially have a look at those. We can can come up with a general rule
(style guide) for those things, and I can fix that. Or you merge it and we fix
remaining style issues later.

(As a side effect I found a fix for #60)

### DIFF
--- a/src/account_manager.hh
+++ b/src/account_manager.hh
@@ -6,8 +6,6 @@
 # include "proto.hh"
 # include "utils/address.hh"
 
-using namespace std;
-
 namespace Astroid {
   class Account {
     public:
@@ -21,7 +19,7 @@ namespace Astroid {
 
       bool save_sent;
       ustring save_sent_to;
-      vector<ustring> additional_sent_tags;
+      std::vector<ustring> additional_sent_tags;
       ustring save_drafts_to;
 
       ustring full_address ();
@@ -34,7 +32,7 @@ namespace Astroid {
       AccountManager ();
       ~AccountManager ();
 
-      vector<Account> accounts;
+      std::vector<Account> accounts;
       int default_account;
       Account * get_account_for_address (ustring);
       Account * get_account_for_address (Address);
--- a/src/actions/action.hh
+++ b/src/actions/action.hh
@@ -1,19 +1,17 @@
 # pragma once
 
 # include <glibmm.h>
+# include <vector>
 
-# include "astroid.hh"
 # include "proto.hh"
-
-using namespace std;
 
 namespace Astroid {
   class Action : public Glib::Object {
     public:
       Action (refptr<NotmuchThread>);
-      Action (vector<refptr<NotmuchThread>>);
+      Action (std::vector<refptr<NotmuchThread>>);
 
-      vector<refptr<NotmuchThread>> threads;
+      std::vector<refptr<NotmuchThread>> threads;
 
       virtual bool doit (Db *) = 0;
       virtual bool undo (Db *) = 0;
--- a/src/actions/action_manager.hh
+++ b/src/actions/action_manager.hh
@@ -5,17 +5,14 @@
 # include <sigc++/sigc++.h>
 # include <glibmm/threads.h>
 
-# include "astroid.hh"
 # include "proto.hh"
-
-using namespace std;
 
 namespace Astroid {
   class ActionManager {
     public:
       ActionManager ();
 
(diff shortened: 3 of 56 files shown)


gauteh:
Great! I knew my bad habits would come back and bite me. I definitely want some coding style guidelines, I had some vague stuff in [Contributing](https://github.com/gauteh/astroid/wiki/Contributing). I would happily accepts suggestions, or just  go ahead and do it and then we can tweak it afterwards.

Other contributions and improvements are very much appreciated. Cleaner, and better structured, code is safer, less bug-prone, easier to work with and easier to contribute to.

(merged).

gauteh:
In e.g.: theme.cc there is still a using boost::filesystem while there
is a using bfs::path further down, is this intentional?

ff2000:
Excerpts from Gaute Hope's message of Januar 18, 2016 3:54 :
> In e.g.: theme.cc there is still a using boost::filesystem while there
> is a using bfs::path further down, is this intentional?
I think there are still plenty "using namespace" in implementation files
around. This one looks like I have missed.

I also would have liked to discuss some general rules what to allow and
what not.
I think it is also good to prevent `using namespace std` in the scope of
a translation unit. I also think most of the time it doesn't hurt to
type `std::`. If it gets tedious (like all those `endl`) IMHO pulling in
one name with `using std::endl` is OK.
In the tests I pulled in some symbols in the global namespace, I am not
so happy with that, but that are 'just' tests...

Boost is something special because of the long and nested namespace
names. IMHO aliasing is the best way to deal with that (even with a
semantic completion engine in place), and doing it in globally doesn't
hurt.

That's just my opinion. Astroid is your project and if you think this is
too pedantic I'm fine with it :) But if we really want to get some sort
of consistency it's better to know what to do, because otherwise the
mess AFTER style changes might be bigger than before ;)

gauteh:
ff2000 writes on January 18, 2016 16:47:
> Excerpts from Gaute Hope's message of Januar 18, 2016 3:54 :
>> In e.g.: theme.cc there is still a using boost::filesystem while there
>> is a using bfs::path further down, is this intentional?
> I think there are still plenty "using namespace" in implementation files
> around. This one looks like I have missed.

Ok, just wondering if it fit into the style changes you were aiming for.
 
> I also would have liked to discuss some general rules what to allow and
> what not.

Definetely, if there are some stuff we could borrow that would be good.
If you have suggestions I would appreciate it, I am not a fan of making
it overly expressive though, that just makes the code a pain to write
and read. Having rules, and as soon as possible would be great!

> I think it is also good to prevent `using namespace std` in the scope of
> a translation unit. I also think most of the time it doesn't hurt to
> type `std::`. If it gets tedious (like all those `endl`) IMHO pulling in
> one name with `using std::endl` is OK.

Yup, agreed. I guess it is only std::endl that it would make sense to
pull in globally. Also, what do you think about putting `using
std::find_if` etc in the functions that need them, it feels tedious to
keep writing std:: for every single one.

As you probably noticed I got tired of writing Glib::ustring, this could
probably be done in a better way.

> In the tests I pulled in some symbols in the global namespace, I am not
> so happy with that, but that are 'just' tests...

:) fine by me.
 
> Boost is something special because of the long and nested namespace
> names. IMHO aliasing is the best way to deal with that (even with a
> semantic completion engine in place), and doing it in globally doesn't
> hurt.

Yeah, `namespace bfs = boost::filesystem;` and similar seems like a good
alternative. 
 
> That's just my opinion. Astroid is your project and if you think this is
> too pedantic I'm fine with it :) But if we really want to get some sort
> of consistency it's better to know what to do, because otherwise the
> mess AFTER style changes might be bigger than before ;)

Agreed. I am sure we can find a good balance. Also, I don't mind sharing
ownership with dedicated contributors. I do not feel particularily
possessive about my coding style, I defintely made some choices on the
pragmatic side of getting a working concept up and going. Please don't
hesitate to suggest changes to my style!

I quickly googled some stuff, but most existing style guildelines seem
very comprehensive (e.g.: http://www.stroustrup.com/JSF-AV-rules.pdf), I
don't expect anyone to read 141 pages before sending a patch to Astroid.

Regards, Gaute

PR #65, opened 2016-01-18
Speed up hacking by minimizing includes of build_config.hh

ff2000:
astroid.cc still needs several seconds to compile, so I was
tempted to create a simple build_info.cc - but that seems to
be a little bit of overkill.
I also thought of removing Astroid::user_agent as it is not
used anywhere, but I don't know your plans for it ;)

### DIFF
--- a/src/astroid.cc
+++ b/src/astroid.cc
@@ -33,6 +33,7 @@ using namespace boost::filesystem;
 /* globally available static instance of the Astroid */
 Astroid::Astroid * (Astroid::astroid);
 Astroid::Log Astroid::log;
+const char* const Astroid::Astroid::version = GIT_DESC;
 
 namespace Astroid {
   Astroid::Astroid () {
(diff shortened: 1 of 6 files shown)


gauteh:
ff2000 writes on January 18, 2016 10:37:
> astroid.cc still needs several seconds to compile, so I was
> tempted to create a simple build_info.cc - but that seems to
> be a little bit of overkill.
> I also thought of removing Astroid::user_agent as it is not
> used anywhere, but I don't know your plans for it ;)
> You can view, comment on, or merge this pull request online at:

This looks good! I use `Astroid::user_agent` in compose_message.cc. Long
build times have been annoying me too.

ff2000:
Oops, missed that use of user_agent, most likely because of all the binary files that came up (I now always use grep -I to ignore binary files... ;))
Is there something you still want to have changed?

gauteh:
ff2000 writes on January 18, 2016 11:29:
> Oops, missed that use of user_agent, most likely because of all the binary files that came up (I now always use grep -I to ignore binary files... ;))
> Is there something you still want to have changed?

Nope, merged! (btw: I usually use 'ack' in place of 'grep' for source
trees).

gauteh:
removing build_config.hh and defining GIT_DESC on the command line just
makes everything rebuild if they are changed. I guess the only way is to
speed it up anymore is to make build_config.cc and a build_config.hh
that is constant. Then only build_config.cc and linking needs to be
redone.

Issue #31, opened 2016-01-03
Unread emails not marked as read in Maildir


hugoroy:
I've noticed that reading an email in astroid removes the notmuch 'unread' tag but does not seem to affect the status of the email as 'new' in maildir. 

Is there something wrong with my notmuch setup or should this be handled by astroid? 

Edit: apparently, astroid does not synchronize tag changes back to maildir flags
with `notmuch_message_tags_to_maildir_flags` according to j4ni on #notmuch on IRC

gauteh:
Do you mean that it still has the tag 'new' or that the unread flag is
not removed on the server? I usually remove the 'new' tag after having
done any tagging on new mail in the notmuch post-new hook or my poll
script.

Neither really is the responsibility of astroid, notmuch can synchronize
some of the tags to maildir flags [0]. These will normally show up on
the server side as well. There are some configuration flags for notmuch
to set that up.

[0] https://notmuchmail.org/special-tags/

Hugo Roy writes on January 3, 2016 16:33:

hugoroy:
No, I remove the notmuch 'new' tag is removed with a `post-new` hook. What I am concerned with are emails which are tagged as "unread" and which are, well, unread.

In Maildir, such emails will be in the "new" directory.

When I open and read such emails, astroid will correctly remove the "unread" tag but this isn't synced with the Maildir "new" directory, so when I sync this with the server (with offlineimap) the read/unread status isn't changed.


As the link suggests, notmuch can take care of "Synchronization between tags and maildir flags". The page says:

> The above tags are recognized by the lib, and synchronized with the corresponding maildir flags through the following lib function calls:
> 
> ```
>    notmuch_message_maildir_flags_to_tags()
>    notmuch_message_tags_to_maildir_flags()
> ```

gauteh:
Hugo Roy writes on January 3, 2016 21:31:
> No, I remove the notmuch 'new' tag is removed with a `post-new` hook. What I am concerned with are emails which are tagged as "unread" and which are, well, unread.
> 
> In Maildir, such emails will be in the "new" directory.
> 
> When I open and read such emails, astroid will correctly remove the "unread" tag but this isn't synced with the Maildir "new" directory, so when I sync this with the server (with offlineimap) the read/unread status isn't changed.
> 
> 
> As the link suggests, notmuch can take care of "Synchronization between tags and maildir flags". The page says:
> 
[...]

PR #51, opened 2016-01-10
Sent tag configurable

gauteh:

PR #74, opened 2016-01-23
[WIP] Refactor config access

ff2000:
This will come in handy for reading ~/.notmuch-config
without interfering astroid-config.

### DIFF
--- a/src/account_manager.cc
+++ b/src/account_manager.cc
@@ -18,7 +18,7 @@ namespace Astroid {
     AccountManager::AccountManager () {
     log << info << "ac: initializing accounts.." << endl;
 
-    ptree apt = astroid->config->config.get_child ("accounts");
+    ptree apt = astroid->config ("accounts");
 
     for (const auto &kv : apt) {
       Account * a = new Account ();
--- a/src/astroid.cc
+++ b/src/astroid.cc
@@ -233,7 +233,11 @@ namespace Astroid {
   }
 
   const boost::property_tree::ptree& Astroid::config (const std::string& id) const {
-    return m_config->config.get_child(id);
+    if (id == "astroid.notmuch") {
+      return m_config->notmuch_config;
+    } else {
+      return m_config->config.get_child(id);
+    }
   }
 
   const StandardPaths& Astroid::standard_paths() const {
--- a/src/astroid.cc
+++ b/src/astroid.cc
@@ -110,17 +110,17 @@ namespace Astroid {
           exit (1);
         }
 
-        ncnf.config_file = path(cnf);
+        ncnf.std_paths.config_file = path(cnf);
 
       } else {
         /* use default */
-        if (exists(ncnf.config_file)) {
-          log << error << "the config file: " << ncnf.config_file.c_str() << " already exists." << endl;
+        if (exists(ncnf.std_paths.config_file)) {
+          log << error << "the config file: " << ncnf.std_paths.config_file.c_str() << " already exists." << endl;
           exit (1);
         }
       }
 
-      log << info << "writing default config to: " << ncnf.config_file.c_str() << endl;
+      log << info << "writing default config to: " << ncnf.std_paths.config_file.c_str() << endl;
       ncnf.load_config (true);
 
       exit (0);
@@ -192,17 +192,17 @@ namespace Astroid {
       }
 
       log << "astroid: loading config: " << vm["config"].as<ustring>().c_str() << endl;
-      config = new Config (vm["config"].as<ustring>().c_str());
+      m_config = new Config (vm["config"].as<ustring>().c_str());
     } else {
       if (test_config) {
-        config = new Config (true);
+        m_config = new Config (true);
       } else {
-        config = new Config ();
+        m_config = new Config ();
       }
     }
 
     /* output db location */
-    ustring db_path = ustring (config->config.get<string> ("astroid.notmuch.db"));
+    ustring db_path = ustring (config("astroid.notmuch").get<string> ("db"));
     log << info << "notmuch db: " << db_path << endl;
 
     /* set up static classes */
@@ -232,8 +232,16 @@ namespace Astroid {
     return 0;
   }
 
+  const boost::property_tree::ptree& Astroid::config (const std::string& id) const {
+    return m_config->config.get_child(id);
+  }
+
+  const StandardPaths& Astroid::standard_paths() const {
+    return m_config->std_paths;
+  }
+
   void Astroid::main_test () {
-    config = new Config (true);
+    m_config = new Config (true);
 
     /* set up static classes */
     Date::init ();
@@ -253,11 +261,11 @@ namespace Astroid {
 
   Astroid::~Astroid () {
     /* clean up and exit */
-    if (accounts != NULL) delete accounts;
+    delete accounts;
     //if (contacts != NULL) delete contacts;
-    if (config != NULL) delete config;
-    if (poll != NULL) delete poll;
-    if (global_actions != NULL) delete global_actions;
+    delete m_config;
+    delete poll;
+    delete global_actions;
 
     log << info << "astroid: goodbye!" << endl;
     log.del_out_stream (&cout);
@@ -272,7 +280,7 @@ namespace Astroid {
     MainWindow * mw = new MainWindow (); // is freed / destroyed by application
 
     if (open_defaults) {
-      ptree qpt = config->config.get_child ("startup.queries");
+      ptree qpt = config ("startup.queries");
 
       for (const auto &kv : qpt) {
         ustring name = kv.first;
--- a/src/astroid.cc
+++ b/src/astroid.cc
@@ -202,7 +202,7 @@ namespace Astroid {
     }
 
     /* output db location */
-    ustring db_path = ustring (config("astroid.notmuch").get<string> ("db"));
+    ustring db_path = ustring (notmuch_config().get<string> ("database.path"));
     log << info << "notmuch db: " << db_path << endl;
 
     /* set up static classes */
@@ -233,11 +233,11 @@ namespace Astroid {
   }
 
   const boost::property_tree::ptree& Astroid::config (const std::string& id) const {
-    if (id == "astroid.notmuch") {
-      return m_config->notmuch_config;
-    } else {
-      return m_config->config.get_child(id);
-    }
+    return m_config->config.get_child(id);
+  }
+
+  const boost::property_tree::ptree& Astroid::notmuch_config () const {
+    return m_config->notmuch_config;
   }
 
   const StandardPaths& Astroid::standard_paths() const {
--- a/src/astroid.hh
+++ b/src/astroid.hh
@@ -40,9 +40,10 @@ namespace Astroid {
 
       MainWindow * open_new_window (bool open_defaults = true);
 
-    private:
-      /* config */
+    protected:
       Config * m_config;
+
+    private:
       bool activated = false;
       void on_signal_activate ();
       void on_mailto_activate (const Glib::VariantBase &);
--- a/src/astroid.hh
+++ b/src/astroid.hh
@@ -19,6 +19,7 @@ namespace Astroid {
       void main_test ();
 
       const boost::property_tree::ptree& config (const std::string& path=std::string()) const;
+      const boost::property_tree::ptree& notmuch_config () const;
       const StandardPaths& standard_paths() const;
 
       refptr<Gtk::Application> app;
--- a/src/astroid.hh
+++ b/src/astroid.hh
@@ -1,6 +1,9 @@
 # pragma once
 
 # include <vector>
+# include <string>
+
+# include <boost/property_tree/ptree.hpp>
 
 # include <gtkmm.h>
 # include <glibmm.h>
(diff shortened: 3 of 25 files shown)


ff2000:
As the title says - work in progress. So don't merge in this state.

ff2000:
Reading and merging of ~/.notmuch-config should work now, at least concerning the current features.
Some commits can be squashed, but I think the one with reading and warning about deprecated values should be kept separate, so we can simply revert that change in case astroid reaches v0.7 (or whichever version will remove the config options).

I hope I will find the time tomorrow to add support for maildir.synchronize_flags.

ff2000:
I now think renaming the keys to access notmuch-config is not really a good idea, because new contributors might get confused as they would expect other keys. Also debugging might get somewhat harder and bugs might slip in. [database.path instead of some config("astroid.notmuch").db]
I think I will also add a separate function "Astroid::notmuch_config()" to make things even more clear.
@gauteh Take your time and think about the changes as soon as you are AFAFK ;)

ff2000:
@gauteh The last commit implements tag_to_maildirflags synchronization. If you want I can create another PR for that - I just wanted to see if everything comes together and fix #31 ;)

ff2000:
What remains:
- Handling of deprecated values: Is it really needed? Or is it enough to tell the user in the release notes to set "astroid.notmuch_config" (or astroid.notmuch.config"?) and use that for db-path and excluded tags - and completely ignore the other notmuch-related settings in astroid-config?
- Can we remove "astroid.notmuch.[db, excluded_tags]" from default_config? I think we can and should.
- Should astroid.config.version be incremented?
- Should "astroid.notmuch.sent_tags" stay where it is? Now that Astroid::notmuch_config() is its own function I think it can. But I wondered if this could be placed in the AccountManager and give the user the possibility to tag sent messages depending on the account. That way one could automatically add a tag for "work" or "family" when using dedicated accounts.

(The last point again could be done in a seperate PR - it just popped up while digging through the code)

gauteh:
I think this looks very promising! I will have to have a more through
look at the config access later, but I like the new approach. I agree
that notmuch_config should be accessed with a separate function (or at
least be clearly separable from the original config).

Some comment below:

ff2000 writes on January 25, 2016 9:24:
> What remains:
> - Handling of deprecated values: Is it really needed? Or is it enough to tell the user in the release notes to set "astroid.notmuch_config" (or astroid.notmuch.config"?) and use that for db-path and excluded tags - and completely ignore the other notmuch-related settings in astroid-config?

Yes, I don't think we need to be too careful about non-critical issues.

> - Can we remove "astroid.notmuch.[db, excluded_tags]" from default_config? I think we can and should.

Yes.

> - Should astroid.config.version be incremented?

Yes, at least for support issues (we can ask the user which version he
runs).

> - Should "astroid.notmuch.sent_tags" stay where it is? Now that Astroid::notmuch_config() is its own function I think it can. But I wondered if this could be placed in the AccountManager and give the user the possibility to tag sent messages depending on the account. That way one could automatically add a tag for "work" or "family" when using dedicated accounts.

I always thought that the .notmuch. sub-section was a bit out of place,
perhaps sent_tags should be under edit_message? Perhaps we can just
remove the whole notmuch section after adding notmuch_config()?

I added support for per-account sent-tags not long ago in #51 :)

ff2000:
Excerpts from Gaute Hope's message of Januar 25, 2016 9:53 :
> I think this looks very promising! I will have to have a more through
> look at the config access later, but I like the new approach. I agree
> that notmuch_config should be accessed with a separate function (or at
> least be clearly separable from the original config).
> 
> Some comment below:
> 
> ff2000 writes on January 25, 2016 9:24:
>> What remains:
>> - Handling of deprecated values: Is it really needed? Or is it enough to tell the user in the release notes to set "astroid.notmuch_config" (or astroid.notmuch.config"?) and use that for db-path and excluded tags - and completely ignore the other notmuch-related settings in astroid-config?
> 
> Yes, I don't think we need to be too careful about non-critical issues.

Ok, (manually) reverted accessing those deprecated values

> 
>> - Can we remove "astroid.notmuch.[db, excluded_tags]" from default_config? I think we can and should.
> 
> Yes.

Done.

> 
>> - Should astroid.config.version be incremented?
> 
> Yes, at least for support issues (we can ask the user which version he
> runs).

Done, incremented to 3.

> 
>> - Should "astroid.notmuch.sent_tags" stay where it is? Now that Astroid::notmuch_config() is its own function I think it can. But I wondered if this could be placed in the AccountManager and give the user the possibility to tag sent messages depending on the account. That way one could automatically add a tag for "work" or "family" when using dedicated accounts.
> 
> I always thought that the .notmuch. sub-section was a bit out of place,
> perhaps sent_tags should be under edit_message? Perhaps we can just
> remove the whole notmuch section after adding notmuch_config()?

I did not do that, yet. There is a "mail" section that already
configures sending-preferences - mail.sent_tags sunds ok, too.
Probably a new section "account_defaults" would make sence? There we
could also store (fallback)defaults for sendmail, save_sent,
save_sent_to, save_drafts_to, so users like me wouldn' have to dupe
those settings for each account. ;)

But I'm fine with edit_message.sent_tags, too.

In any case we need to make sure that old settings get moved to the new
location - which should be doable with a small config_version_check.

> 
> I added support for per-account sent-tags not long ago in #51 :)

Thx, completely missed that :)

===

I did not yet squash any commits - Do you want to have several dedicated
commits? Or is it OK to squash it into one commit?

gauteh:
ff2000 writes on January 27, 2016 10:55:
> [...]
>
> Done.

great!

>>> - Should "astroid.notmuch.sent_tags" stay where it is? Now that Astroid::notmuch_config() is its own function I think it can. But I wondered if this could be placed in the AccountManager and give the user the possibility to tag sent messages depending on the account. That way one could automatically add a tag for "work" or "family" when using dedicated accounts.
>>
>> I always thought that the .notmuch. sub-section was a bit out of place,
>> perhaps sent_tags should be under edit_message? Perhaps we can just
>> remove the whole notmuch section after adding notmuch_config()?
>
> I did not do that, yet. There is a "mail" section that already
> configures sending-preferences - mail.sent_tags sunds ok, too.
> Probably a new section "account_defaults" would make sence? There we
> could also store (fallback)defaults for sendmail, save_sent,
> save_sent_to, save_drafts_to, so users like me wouldn' have to dupe
> those settings for each account. ;)

I guess all of those could go in the 'mail' section too? However, I am a
little bit reluctant to add fallbacks like that. It is after all not
that much of a hassle to copy-paste your first account config. It feels
safer and less complicated to rather fail than fallback to something un-inteded.


> But I'm fine with edit_message.sent_tags, too.

'mail' makes more sense to me. then 'dryrun_sending' can go in there
too. agreed?

> In any case we need to make sure that old settings get moved to the new
> location - which should be doable with a small config_version_check.

Yes, I am afraid this will become a pain to maintain. But hopefully it
the config file will stabilize soon, let's make sure we get it right now
so that we don't end up with more checks.

> ===
>
> I did not yet squash any commits - Do you want to have several dedicated
> commits? Or is it OK to squash it into one commit?

Several dedicated commits is better! Once the branch matures we can fix
up the reverts and minor improvements and squash them a bit together.
You can do this and push --force if you want (once it is basically ready
for merging), or I can try to make sense of it before I merge.

ff2000:
> I guess all of those could go in the 'mail' section too? However, I am a
> little bit reluctant to add fallbacks like that. It is after all not
> that much of a hassle to copy-paste your first account config. It feels
> safer and less complicated to rather fail than fallback to something un-inteded.

Agreed. Let's go the safe way and keep it as it is.

>> But I'm fine with edit_message.sent_tags, too.
> 
> 'mail' makes more sense to me. then 'dryrun_sending' can go in there
> too. agreed?

Agreed

>> In any case we need to make sure that old settings get moved to the new
>> location - which should be doable with a small config_version_check.
> 
> Yes, I am afraid this will become a pain to maintain. But hopefully it
> the config file will stabilize soon, let's make sure we get it right now
> so that we don't end up with more checks.

Could be done in a seperate function, so that future config version
changes can be migrated step-by-step to the most recent version.

> Several dedicated commits is better! Once the branch matures we can fix
> up the reverts and minor improvements and squash them a bit together.
> You can do this and push --force if you want (once it is basically ready
> for merging), or I can try to make sense of it before I merge.

Would be great if you could do that. I'm still not recovered and have a
hard time to even answer E-Mails (as you might have realised ;))

gauteh:
ff2000 writes on January 29, 2016 9:59:
>>> In any case we need to make sure that old settings get moved to the new
>>> location - which should be doable with a small config_version_check.
>>
>> Yes, I am afraid this will become a pain to maintain. But hopefully it
>> the config file will stabilize soon, let's make sure we get it right now
>> so that we don't end up with more checks.
>
> Could be done in a seperate function, so that future config version
> changes can be migrated step-by-step to the most recent version.

I am sure you already noticed `[Config::check_config (...)](https://github.com/gauteh/astroid/blob/master/src/config.cc#L272)`.

>> Several dedicated commits is better! Once the branch matures we can fix
>> up the reverts and minor improvements and squash them a bit together.
>> You can do this and push --force if you want (once it is basically ready
>> for merging), or I can try to make sense of it before I merge.
>
> Would be great if you could do that. I'm still not recovered and have a
> hard time to even answer E-Mails (as you might have realised ;))

No worries, it's supposed to be fun. I am really busy the next days too.

gauteh:
Gaute Hope writes on January 29, 2016 12:35:
> ff2000 writes on January 29, 2016 9:59:
>>>> In any case we need to make sure that old settings get moved to the new
>>>> location - which should be doable with a small config_version_check.
>>>
>>> Yes, I am afraid this will become a pain to maintain. But hopefully it
>>> the config file will stabilize soon, let's make sure we get it right now
>>> so that we don't end up with more checks.
>>
>> Could be done in a seperate function, so that future config version
>> changes can be migrated step-by-step to the most recent version.
> 
> I am sure you already noticed `[Config::check_config (...)](https://github.com/gauteh/astroid/blob/master/src/config.cc#L272)`.
> 
>>> Several dedicated commits is better! Once the branch matures we can fix
>>> up the reverts and minor improvements and squash them a bit together.
>>> You can do this and push --force if you want (once it is basically ready
>>> for merging), or I can try to make sense of it before I merge.
>>
>> Would be great if you could do that. I'm still not recovered and have a
>> hard time to even answer E-Mails (as you might have realised ;))
> 
> No worries, it's supposed to be fun. I am really busy the next days too.
> 
> 

You can check out my changes on:
https://github.com/gauteh/astroid/tree/refactor-config-gaute , in the
end I went back to warn about deprecation, but not fix it. I also found
a few things that you missed and rebased it to current master, let me
know if I have been using it wrongly. I was planning on squashing out
some of the commits that you seem to have discarded in later commits.

gauteh:
As it is now, any error messages concerning the config is drowned in
everything else. Need to remove most of the debug output. Also, perhaps
make a script that looks for calls to `register_key` and `register_run`
to find keys and hooks without needing the user to scan through the
terminal output. I do not run astroid in the terminal, and it shouldn't
be necessary.

Perhaps a log window should be opened automatically every-time there is
an error? Or the output could be cached for a say 1000 lines or
something.

ff2000:
But there aren't any specific errors this PR causes for you?

I also thought about that. Especially when there are warnings/errors at
startup. Don't know if it's overkill to make the LogLevel configurable,
so that pedantics could also get notified about Warnings? Would be
especially great if also the log window/console output could ignore
everything lower than MinimumLogLevel to minimize [ DEBUG ] spamming ;)
I think an ordinary user might not be interested always in DEBUG
messages. A cmdline-switch to overwrite config settings for debug
purposes would make sense, then. (--log-level=debug)

Concerning notification about errors:
There is that nice bar that pops up in the bottom of the window to query
for user input. Couldn't that be used?
`There were 12 warnings and 3 errors. Open log window? (yes no)`
IMHO would be more user friendly than just forcing focus on a different
mode...
Caching makes sense, but probably using a timer to notify the user would
be better?
And if the log window already is opened, one could set notify-state on
the tab and fade in an error/warning/... icon on the right side of the
tab bar (like when sending a message)

Excerpts from Gaute Hope's message of Januar 30, 2016 11:06 :

gauteh:
ff2000 writes on January 30, 2016 12:26:
> But there aren't any specific errors this PR causes for you?
> 
> I also thought about that. Especially when there are warnings/errors at
> startup. Don't know if it's overkill to make the LogLevel configurable,
> so that pedantics could also get notified about Warnings? Would be
> especially great if also the log window/console output could ignore
> everything lower than MinimumLogLevel to minimize [ DEBUG ] spamming ;)
> I think an ordinary user might not be interested always in DEBUG
> messages. A cmdline-switch to overwrite config settings for debug
> purposes would make sense, then. (--log-level=debug)

Yes, both good!

> Concerning notification about errors:
> There is that nice bar that pops up in the bottom of the window to query
> for user input. Couldn't that be used?
> `There were 12 warnings and 3 errors. Open log window? (yes no)`
> IMHO would be more user friendly than just forcing focus on a different
> mode...
> Caching makes sense, but probably using a timer to notify the user would
> be better?
> And if the log window already is opened, one could set notify-state on
> the tab and fade in an error/warning/... icon on the right side of the
> tab bar (like when sending a message)

We could cache serious errors, or cache them untill the user opens a
window. Especially errors occuring before the mainwindow is opened needs
to be cached. Perhaps just a info bar, or icon, and then fade it out
after a little while (but not the icon untill the log view is opened). I
like it better if the user dont have to interact with the warning/error
and can just ignore it.

Could maybe put the wrong db path stuff in here too?

gauteh:
ff2000 writes on January 30, 2016 12:26:
> But there aren't any specific errors this PR causes for you?

No, it seems to be working fine. Been running it (with my minor
modifications) for a litte while now.

gauteh:
Merged. So users have to know about:

* astroid.notmuch_config points to the correct ~/.notmuch-config
* astroid.notmuch.exclude_tags must be updated in ~/.notmuch-config
  (implemented after last release).
* astroid.notmuch.sent_tags have been moved to astroid.mail.sent_tags

PR #76, opened 2016-01-26
Run: add run shell-command hooks, starting out with thread_index.run

gauteh:
This implements #29 and allows custom shell commands to be bound to keys in astroid. For now there is only `thread_index.run`. This allows you to have a shell command be run with the `thread` as an argument. 

The run-hooks are configured through the keybindings file, the most rudimentary example which allows you to safely get the hang of it is to add a line like:

~/.config/astroid/keybindings:
```
thread_index.run(echo %1)=w
```

if you press `w` in the Thread Index the command `echo %1` will be run with `%1` replaced with the `thread`-id from the currently selected thread. 

A more useful example is to set up keybindings to toggle custom tags, first create a script that can toggle its given tag. In this example the `queue` tag:

~/.config/astroid/hooks/toggle (you can place this anywhere):
```sh
#! /usr/bin/bash
#
# get a tag as first argument and thread id as second argument
#

# check if we have tag
if [[ $(notmuch search thread:$2 and tag:$1) ]]; then
  echo "removing tag: $1 from thread:$2"
  notmuch tag -$1 thread:$2
else
  echo "adding tag: $1 from thread:$2"
  notmuch tag +$1 thread:$2
fi
``` 

then create a keybinding in:
~/.config/astroid/keybindings:
```sh
# toggle queue
thread_index.run(/home/gaute/.config/astroid/hooks/toggle queue %1)=w

# toggle todo
thread_index.run(/home/gaute/.config/astroid/hooks/toggle todo %1)=M-t
```

you can add several lines for different tags. A successful exit status (0) will trigger a refresh of the thread in astroid ensuring it is updated in all thread indexes. The command is passed to: [`Glib::spawn_command_line_sync`](https://developer.gnome.org/glibmm/stable/group__Spawn.html#ga75961831b4dd3979bb8ab508ee3b3de7). The argument (`thread`-id) is substituted into the string with [`ustring::compose()`](https://developer.gnome.org/glibmm/stable/classGlib_1_1ustring.html#a18e1242bc0ad8a961a28fb2198392258). Future run-hooks for other context may use more arguments, or substitute a list of `thread`-ids into the placeholder.

It would be nice to provide a way to reverse/undo the action. Perhaps being able to register a run-hook for the reverse or provide a flag for the reverse action.

### DIFF
--- a/src/config.cc
+++ b/src/config.cc
@@ -50,7 +50,7 @@ namespace Astroid {
 
       home = cur_path / path("test/test_home");
 
-      log << LogLevel::test << "cf: using home directory: " << home.c_str () << endl;
+      log << LogLevel::test << "cf: using home and config_dir directory: " << home.c_str () << endl;
 
     } else {
       char * home_c = getenv ("HOME");
@@ -64,11 +64,15 @@ namespace Astroid {
     }
 
     /* default config */
-    char * config_home = getenv ("XDG_CONFIG_HOME");
-    if (config_home == NULL) {
-      config_dir = home / path(".config/astroid");
+    if (test) {
+      config_dir = home;
     } else {
-      config_dir = path(config_home) / path("astroid");
+      char * config_home = getenv ("XDG_CONFIG_HOME");
+      if (config_home == NULL) {
+        config_dir = home / path(".config/astroid");
+      } else {
+        config_dir = path(config_home) / path("astroid");
+      }
     }
 
     /* default data */
--- a/src/modes/keybindings.cc
+++ b/src/modes/keybindings.cc
@@ -43,10 +43,13 @@ namespace Astroid {
 
         /* the bindings file has the format:
          *
-         * ``` thread_index.next_thread=j thread_index.next_thread=Down
+         * ```
+         * thread_index.next_thread=j
+         * thread_index.next_thread=Down
          * thread_index.label=C-j
          *
-         * # thread_ ```
+         * # thread_
+         * ```
          *
          * blank lines, or lines starting with # are ignored. a keybinding can
          * be listed several times, in which case they will be interpreted as
@@ -136,6 +139,7 @@ namespace Astroid {
 
             k.name = name;
             k.allow_duplicate_name = true;
+            k.userdefined = true;
 
             user_run_bindings.push_back (std::make_pair (k, target));
 
--- a/src/modes/keybindings.cc
+++ b/src/modes/keybindings.cc
@@ -98,16 +98,19 @@ namespace Astroid {
             fnd = line.find ("(", fnd);
 
             if (fnd == std::string::npos) {
-              ustring erru = "invalid 'run'-keyspec";
-              log << error << "ky: " << erru << endl;
-              throw keyspec_error (erru.c_str ());
+              log << error <<  "ky: invalid 'run'-specification: no '('" << endl;
+              continue;
             }
 
             std::size_t rfnd = line.rfind ("=");
             if (rfnd == std::string::npos) {
-              ustring erru = "invalid 'run'-keyspec";
-              log << error << "ky: " << erru << endl;
-              throw keyspec_error (erru.c_str ());
+              log << error << "ky: invalid 'run'-specification: no '='" << endl;
+              continue;
+            }
+
+            if (rfnd < fnd) {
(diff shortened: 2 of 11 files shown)


ff2000:
Will test this later.
What I wonder: Wouldn't it be more user-friendly if not the full path would be required but only the cmd?
thread_index.run(toggle queue %1)=w
or probably
thread_index.run(hooks/toggle queue %1)=w
We can concatenate paths and test if the file exists and run it, otherwise assume it's in $PATH and run it as specified.

gauteh:
ff2000 writes on January 27, 2016 11:11:
> Will test this later.
> What I wonder: Wouldn't it be more user-friendly if not the full path would be required but only the cmd?
> thread_index.run(toggle queue %1)=w
> or probably
> thread_index.run(hooks/toggle queue %1)=w
> We can concatenate paths and test if the file exists and run it, otherwise assume it's in $PATH and run it as specified.

Yeah I wish that was included in Glib, but we don't need to assume the
script is located in ~/.config/astroid/hooks. It could be anywhere.
Perhaps there is a shell-expand that works with Glibmm::. Trying to
manually expand '~' (which is the most critical) could easily lead to
weird things since the command might expect it too. E.g. make a command
to open a terminal and apply a patch to a git repo could contain stuff
like HEAD~1.

gauteh:
Maybe http://man7.org/linux/man-pages/man3/wordexp.3.html

ff2000:
Excerpts from Gaute Hope's message of Januar 27, 2016 11:16 :
> ff2000 writes on January 27, 2016 11:11:
>> Will test this later.
>> What I wonder: Wouldn't it be more user-friendly if not the full path would be required but only the cmd?
>> thread_index.run(toggle queue %1)=w
>> or probably
>> thread_index.run(hooks/toggle queue %1)=w
>> We can concatenate paths and test if the file exists and run it, otherwise assume it's in $PATH and run it as specified.
> 
> Yeah I wish that was included in Glib, but we don't need to assume the
> script is located in ~/.config/astroid/hooks. It could be anywhere.
> Perhaps there is a shell-expand that works with Glibmm::. Trying to
> manually expand '~' (which is the most critical) could easily lead to
> weird things since the command might expect it too. E.g. make a command
> to open a terminal and apply a patch to a git repo could contain stuff
> like HEAD~1.

Probably I wasn't clear enough:
I think most of the time the user will run small scripts just written
for astroid. We can document (and will) that those script should go into
~/.config/astroid/hooks.
In src/cmd.cc: Cmd::run we then take the string (toggle), append
it to Config::config_dir + "/hooks" and (if it exists) pass that to
Glibmm::spawn. To be save to get what the user wants we could add a
prefix: "hook::toggle" to make clear: this is a script in
astroid/config/hooks.
thread_index.run(hooks::toggle queue %1)=w
(probably "::" is a better seperator than "/" in my previous post)

The outcome should be the same but the user doesn't have to specify the
full path (in case he places his scripts in the dedicated hooks-dir)

gauteh:
Yeah, that works. `hooks::` seems like a good choice. Then you can skip the prefix if you want to run a regular command.

gauteh:
One concern; perhaps we should prefix everything with `sh::` so that we later can implement a few commands like `search` or `reply`? Once `astroid -s` works you can make a hook that launches it, but some thing are better to do right away.

gauteh:
Keyboard macros might be produced like this: https://mail.gnome.org/archives/gtk-app-devel-list/2004-July/msg00016.html this could be used to implement regular macros.

ff2000:
If you want to do general scripting you might consider integrating lua -
probably more lightweight than python or ruby.
For something really basic you might get inspiration from
[kakoune](https://github.com/mawww/kakoune) - for examples look at "rc"
subdir.

In general I think it might be better to implement basic stuff like
"reply" and "search" in C++, as it is something the user might expect
out-of-the-box.

Hopefully in some days I will better get what you are heading for...

gauteh:
I think I will just merge what I got now, if we later want to make
integrated commands I'll either create a separate target (other than
`run`) or prefix integrated stuff with `astroid::`.

Agreed, basic stuff should just be implemented.

hugoroy:
This looks interesting but I don't think I understand it clearly enough to write a wiki page about it yet :-)

PR #87, opened 2016-02-22
Plaindate

emdete:
i added a way to always use the date formatting setting from same_year/diff_year without any intelligence to calculate "yesterday" or the like.

### DIFF
--- a/src/utils/date_utils.cc
+++ b/src/utils/date_utils.cc
@@ -56,39 +56,52 @@ namespace Astroid {
     CoarseDate cd = coarse_date (t);
 
     ustring fmt;
-    switch (cd) {
-      case CoarseDate::NOW:
-        return "Now";
+	if (clock_format == ClockFormat::YEAR) {
+		switch (cd) {
+		case CoarseDate::YEARS:
+		case CoarseDate::FUTURE:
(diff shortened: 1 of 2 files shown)


gauteh:
Patch looks good. Is it logical that plain_date gets its setting from
same_year or diff_year? Would it be better if it just got the format
from one setting? Maybe even clock_format?

So that if clock_format is none of the special supported it will use the
format from that setting?

(we should document these settings in the wiki anyway!)

Regards, Gaute

src/utils/date_utils.cc
```diff
@@ -134,6 +146,8 @@ namespace Astroid {
       clock_format = ClockFormat::TWENTY_FOUR_HOURS;
     } else if (c_f == "12h") {
       clock_format = ClockFormat::TWELVE_HOURS;
+    } else if (c_f == "plain") {
+      clock_format = ClockFormat::PLAIN;
     } else if (c_f == "local") {
       clock_format = ClockFormat::LOCALE_DEFAULT;
     } else {
```
gauteh: error message should include 'plain' now, unless we go for changes in the other comment.

gauteh:
Merged.

PR #93, opened 2016-02-22
Test bad content part: blank line between separator and Content-Type causes Content-Type not to be read

gauteh:
blank line between separator and Content-Type causes Content-Type not to be read

bad-content-part-id.eml shows original with blank line at line 30, gmime returns no content type here
bad-content-part-id2.eml has blank line removed, and parses as expected.


gauteh:
@emdete

gauteh:
Merged.

PR #95, opened 2016-02-24
ti: use state-less queries: load all threads in the background

gauteh:
keeping notmuch query objects alive for a long while causes issues since
they are quite fragile to outside (or inside database changes). all
internal changes to the database (including polling with notmuch lastmod
functionality) should be caught.

this patch loads all threads in the background while the user interface remains responsive, there are some threading complications that needs particular attention:
- changing threads while loading: in case the thread has not been loaded yet (**deferred**)
- adding / deleting threads which have not been added yet (**deferred**)
- invalidated query while loading: what do we do? restart? defer the action to loading is done (blocking the UI)? (**currently deferring**)

### DIFF
--- a/History.txt
+++ b/History.txt
@@ -13,6 +13,7 @@
 * Gravatar is configurable.
 * #103: Drafts are saved on force close or quit by default.
 * C-v: duplicate and refine query.
+* Use state-less queries in thread-index
 
 == v0.5 / 2016-02-06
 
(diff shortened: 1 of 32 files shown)


gauteh:
https://notmuchmail.org/pipermail/notmuch/2016/021895.html

gauteh:
Actions are now done async, meaning if you do a change that requires writing to the db (at least most common ones) they will wait in a separate thread untill all dbs are closed. Many RO dbs are allowed at the same time, but not if there is any RW dbs open.

PR #97, opened 2016-02-24
Get keys regex

emdete:
rewrite of the parsing part of getting key bindings from cc files using regex

### DIFF
--- a/devel/get_keys.py
+++ b/devel/get_keys.py
@@ -22,12 +22,14 @@ def stripper(s):
 
 def out_binding(multi, key, function, documentation, ):
 	if 'UnboundKey' in key:
-		key = 'no defaults'
+		multi = 'no defaults'
 	else:
-		key  = 'default: ' + stripper(key)
+		key  = stripper(key)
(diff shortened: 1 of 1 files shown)


gauteh:
thanks, this looks good, I think either:

- fix the " more:" to have a comma in front and only appear if there are
  aliases,
- or make new lines as you suggested.

emdete:
done so. so all lines are comments now, all default keys at the end. so you can add / change stuff without loosing doc of the default

gauteh:
M. Dietrich writes on February 24, 2016 19:49:
> done so. so all lines are comments now, all default keys at the end. so you can add / change stuff without loosing doc of the default

looks good, but I think it made more sense to have the default after the
'=' too. then it is easy to see what needs to be replaced, and you can
just uncomment and replace the key when you want to make a modification.

emdete:
done.

gauteh:
M. Dietrich writes on February 25, 2016 13:35:
> done.

Thanks, merged.

PR #112, opened 2016-04-04
#31: synchronize maildir flags

gauteh:
Please test


gauteh:
There's a double free when deleting drafts that I haven't figured out yet...

Issue #113, opened 2016-04-17
Loading too much time for a loooong email


hugoroy:
1. Open a very long email (I'm talking about a 130-page of text in one email)
2. See astroid taking a looong time

Console output:

```
[debug] (18:13:14): theme: loading..
[info ] (18:13:14): re: using user configured resource: /home/hugo/ui/thread-view.html
[info ] (18:13:14): re: using user configured resource: /home/hugo/ui/thread-view.scss
[debug] (18:13:14): tv: testing version: 2
[debug] (18:13:14): tv: testing version: 2
[info ] (18:13:14): theme: processing: /home/hugo/.config/astroid/ui/thread-view.scss
[warn ] (18:13:14): key: + (thread_view.tag_message) already exists in map with name: thread_view.multi, overwriting.
[info ] (18:13:14): tv: load thread: 00000000000206bf
[info ] (18:13:14): db: open db read-only, waiting for lock..
[debug] (18:13:14): db: read-only got lock.
[debug] (18:13:14): db: open time: 0.155 ms.
[info ] (18:13:14): msg: loading mid: CADNfgRcLCHzin2B+HgiZNOTjtP7B5eN92E_7Fm2awawO+zfmjQ@...
[info ] (18:13:14): msg: filename: /home/hugo/Mail/ampoliros/INBOX/cur/1460900188_3.28651.xps,U=20193,FMD5=7e33429f656f1e6e9d79b29c3f82c57e:2,RS
[debug] (18:13:14): chunk (0): content-type: multipart/mixed
[debug] (18:13:14): chunk: multi part
[debug] (18:13:14): chunk: alternative: 0
[debug] (18:13:14): chunk (1): content-type: multipart/alternative
[debug] (18:13:14): chunk: multi part
[debug] (18:13:14): chunk: alternative: 1
[debug] (18:13:14): chunk (2): content-type: text/plain
[debug] (18:13:14): chunk: preferred.
[debug] (18:13:14): chunk: is part (viewable: 1, attachment: 0) 
[debug] (18:13:14): chunk (3): content-type: text/html
[debug] (18:13:14): chunk: is part (viewable: 1, attachment: 0) 
[debug] (18:13:14): chunk: multipart: added sibling
[debug] (18:13:14): chunk: multipart: preferred.
[debug] (18:13:14): chunk: multipart: added sibling
[debug] (18:13:14): chunk: multi part end
[...]

PR #114, opened 2016-04-20
Utilize GMIME_FILTER_HTML_BLOCKQUOTE_CITATIONS

gauteh:
You need a patched gmime:

https://mail.gnome.org/archives/gmime-devel-list/2016-April/msg00004.html
https://github.com/gauteh/gmime/tree/blockquote_html

https://bugzilla.gnome.org/show_bug.cgi?id=765417

Should fix issue #113.


gauteh:
If including the patch in gmime takes too long time we'll just include the custom filter.

Issue #42, opened 2016-01-06
(feature) persistent history of search queries


hugoroy:
When making a new query (with F by default), it doesn't seem to remember past queries. Making this persistent would be useful so that users can simply relaunch a recent query if it's complex enough to bother.

gauteh:
Absolutely! If you feel adventurous, I'd be happy to receive a pull-request :beers: .

hugoroy:
This is unlikely to happen. I know nothing of c plus plus.

If I had some vacation I would give it a try. But I'm not a developer so it's not something I might do by throwing just over a couple of hours at it.

However, if there's anything related to licensing, I could make a pull request (I'm an attorney).

gauteh:
Ok, I appreciate the suggestions, debugging and wikiing. Hopefully,
there won't be any license issues :)

I realize c++ is not easiest to help out with.

ff2000:
Where should those queries be stored? I think it's better to not touch a hand-crafted config file via code, so I would prefer e.g. a seperate simple plain text file where astroid stores those queries.
The other thing is: What actually IS history? Really everything in chronological order? Or is it OK to delete a query when you ran the same right now? (deduplication - I already had discussions about that ;)) I prefer deduplication. If it get's controversial should there be a config switch? (no deduplication; deduplication without reordering; deduplication with reordering - some more options?)

gauteh:
I was looking at notmuch-hello for the emacs interface (https://notmuchmail.org/screenshots/), we could do something similar. Save the previous searches, and allow some to be saved (at some point). I also wanted to create a screen with an overview of tags with the number of messages in them, then the saved searches could go there. Anyway, that is a little more work.

I also think it should not be in the config file, keep say the last 100 in ~/.config/astroid/last_searches (or .cache/astroid/last_searches) and then possibly the saved searches in ~/.config/astroid/saved_searches (or the config file).

The most straightforward would probably be to just put everything in there, without worrying about duplication. If we do, I would prefer de-duplication with reordering.

gauteh:
Search history implemented in #123.

mardukbp:
@hugoroy In the search box you can use the up arrow to reuse recent queries. Hope this helps.

gauteh:
Indeed, this is merged now.

PR #123, opened 2016-04-29
saved searches: show startup queries

gauteh:
#42

Open with `C-f`.

currently only startup queries are shown

The plan is to include:

- [x] saved searches
- [x] update of stats when something happens
- [x] search history is saved, browsable in saved-searches and with Up/Down in search-bar 
- [x] make search history entries savable
- [x] J/K jump several lines
- [x] config for how many histories to show
- [x] possibly keep updated search history list in memory
- [x] make configurable (and default) to show saved searches on startup
- [x] double-click opens query

this is getting pretty close to: https://notmuchmail.org/screenshots/#index1h3

### DIFF
--- a/History.txt
+++ b/History.txt
@@ -6,7 +6,8 @@
 * Signatures: Include per-account signatures (either included or
   attached).
 * Saved searches: Save a search with 'C-S', then browse saved searches
-  and stats with 'C-f'.
+  and stats with 'C-f'. You can also browse search history using Up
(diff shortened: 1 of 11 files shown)


gauteh:
by the way, I changed the defaults to:

- don't show saved searches on startup by default
- only show 15 searches in search-history by default (to avoid spending a lot of time on refreshing search stats everytime something happens)

PR #126, opened 2016-05-02
add ml to: consideration in account detection

emdete:
improve account detection for ML emails. use other headerfields to detect own address.


gauteh:
Thanks. Merged (with some formatting and a simple test-case).

Issue #139, opened 2016-05-18
Colored tags


aureooms:
Would there be an easy way to specify different colors for different tags? Possibly background color with matching text color like in gmail, random would be ok, better if it can be matched with gmail color (I don't know if there is any way to retrieve this). I would find this very useful. Tags displayed in the thread index in monochrome color do not seem to help me to sort through mail in any way.

gauteh:
Aurélien Ooms writes on mai 18, 2016 22:05:
> Would there be an easy way to specify different colors for different tags? Possibly background color with matching text color like in gmail, random would be ok, better if it can be matched with gmail color (I don't know if there is any way to retrieve this). I would find this very useful. Tags displayed in the thread index in monochrome color do not seem to help me to sort through mail in any way.
> 

I think most ways (not involving synchronization) could be implemented
fairly easily - what would the algorithm be? It would be nice with:

- the same tag gets the same color always
- tags next to each other (alphabetically sorted) dont have the same
  color

aureooms:
Via hashing you get the first but not necessarily the second, although getting the same color for two different tags alphabetically close would be rather rare. The background color of the headers in [this example](https://aureooms-research.github.io/pats-game/) are generated exactly with this technique (last 6 hex digit of the hash). You can restrict to nice pastel colors by mixing a random color with white.

gauteh:
Aurélien Ooms writes on mai 18, 2016 22:44:
> Via hashing you get the first but not necessarily the second, although getting the same color for two different tags alphabetically close would be rather rare. The background color of the headers in [this example](https://aureooms-research.github.io/pats-game/) are generated exactly with this technique (last 6 hex digit of the hash). You can restrict to nice pastel colors by mixing a random color with white.
> 

That could work, what exactly is the algorithm that you use there? I
think it would be best to mix it with light gray or something (if that
works).

aureooms:
SHA1

gauteh:
Aurélien Ooms writes on mai 19, 2016 10:02:
> SHA1

I am using MD5, since I had it readily available without extra
dependencies.

Please test #140. 
- Do you think that the base color should be configurable? 
- What should the default be? 
- Should it be weighted? Like I do now
- Should the color mixing be configurable?

aureooms:
I will give it a shot tonight

gauteh:
Aurélien Ooms writes on mai 19, 2016 10:57:
> I will give it a shot tonight
> 

[...]

PR #140, opened 2016-05-19
ti: vary tag color based on md5 hashing

gauteh:
each tag is assigned a color based on the last 6 digits in the md5 sum
of the tag. the color is blended with a base color weighting the base
color with a coefficient of 0.5 to have colors oscilliate around the
base.

each tag will therefore always have the same color, since the last
digits are used it is unlikely that similar tags (and adjacent in the
listing) will have the same color.

the base color is configurable, the same is the weight. by setting the weight to 0 only the base color is used.

#139

### DIFF
--- a/History.txt
+++ b/History.txt
@@ -1,22 +1,35 @@
 ==
 
-* Thread-index: load all threads in the background.
-* Synchronize maildir flags if so configured in .notmuch-config.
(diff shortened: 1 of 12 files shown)


gauteh:
Once agreed on color:

- [ ] color tags in thread-view

Issue #141, opened 2016-05-20
Search for words in thread view


mardukbp:
As far as I can tell, it is currently not possible to search for words when you open a thread. It would be very useful to have this functionality.

gauteh:
Marduk Bolaños writes on mai 20, 2016 22:53:
> As far as I can tell, it is currently not possible to search for words when you open a thread. It would be very useful to have this functionality.
> 

Yes, this has been on my list for-ever: please try #142.

gauteh:
If you are unsure on how to test, have a look at: https://github.com/gauteh/astroid/wiki/Contributing#how-to-test-a-pull-request

mardukbp:
Thanks! How does one move between search results? Like n and N in less.

I guess you are not using Xapian for this feature, because it did not
recognize a search term like "open a thread".

gauteh:
Marduk Bolaños writes on mai 22, 2016 12:17:
> Thanks! How does one move between search results? Like n and N in less.

It says in the pull-request.
 
> I guess you are not using Xapian for this feature, because it did not
> recognize a search term like "open a thread".

No, I use webkit, case-insensitive, but you have to type exactly the
correct phrase.

gauteh:
Merged.

mardukbp:
It only remains to add this command to the help menu.

gauteh:
No. It is automatically added.

mardukbp:
Scratch that. I just saw it. However, there is something wrong with the
key for Cancel current search. It says 65307.

gauteh:
fixed

PR #142, opened 2016-05-21
Thread view search

gauteh:
Search in thread view using `/`.

You can search for text-strings in thread-view using `/` (I opted to not shadow `F`, but you can re-configure this yourself). Then browse the search-results with `N` and `P` (currently shadowing manual poll from thread-view). Stop searching and drop highlighting using `Escape`.

Focusing is a bit of a problem as there does not seem to be a way to get the position of the currently selected text (or the actual text without going through the clipboard). But whenever webkit scrolls the view to a new result it centers it, so when there is a scroll event during searching I focus the centered message, otherwise focus might be a bit off.

When searching all messages are expanded, otherwise hidden matches may be found when iterating over the results.

#141.

### DIFF
--- a/History.txt
+++ b/History.txt
@@ -27,6 +27,9 @@
 * Gravatar is configurable.
 * #103: Drafts are saved on force close or quit by default.
 * C-v: duplicate and refine query.
+* Mark thread moves to next by default.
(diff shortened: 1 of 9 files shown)


hugoroy:
This is *awesome*

PR #162, opened 2016-06-20
make gravatar & avatar-plugin independent

emdete:
if gravatar is configured off its not used, still a plugin may deliver a uri for an avatar which is still used.

### DIFF
--- a/src/modes/thread_view/thread_view.cc
+++ b/src/modes/thread_view/thread_view.cc
@@ -939,13 +939,17 @@ namespace Astroid {
     {
       ustring uri = "";
       auto se = Address(m->sender);
-# ifndef DISABLE_PLUGINS
+# ifdef DISABLE_PLUGINS
+      if (false) {
+# else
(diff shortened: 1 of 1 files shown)


src/modes/thread_view/thread_view.cc
```diff
@@ -936,28 +936,30 @@ namespace Astroid {
...
-
+    {
       ustring uri = "";
+      auto se = Address(m->sender);
 # ifndef DISABLE_PLUGINS
-      if (!plugins->get_avatar_uri (se.email (), Gravatar::DefaultStr[Gravatar::Default::RETRO], 48, uri)) {
+      if (plugins->get_avatar_uri (se.email (), Gravatar::DefaultStr[Gravatar::Default::RETRO], 48, uri)) {
+        ; // all fine, use plugins avatar
+      } else
 # endif
```
gauteh: Looks good, should have `{` here (and at the end then wrapped in #ifndef's)

gauteh:
Thanks, squashed and merged!

Made the following minor changes:
* moved } and else { to the same line
* used (!ustring::empty ()) to check for empty string

PR #167, opened 2016-06-29
fix cutting of quoting-chars

emdete:


gauteh:
I don't think what I proposed on irc would work right either, now it
would not catch '>  >' (more than one space between '>').

Maybe something like:

```
while (start < inptr && *start == '>') {

  start++;

  /* skip next spaces, but only if they are followed by a '>' */
  char * next = start;
  while (next < inptr && *next == ' ') {
    next++;
  }

  if (next < inptr && *next == '>') {
    start = next;
  } else {
    break; // no need to re-evaluate outer while loop
  }
}

```

I am pretty sure 'inptr' points to the next new-line ('\n'), so it
should not be necessary to check for that (or was it 'inend'?).

It still does not handle '>From', but should handle:

```
>> a
> > a
>   > a
```

The last case '>  >' does not seem to be cought by `citation_depth`. So
we could also just say that it is invalid, and that there can be a
maximum of one spaces inbetween the '>'. Then the irc code would work,
and the equivialent of replacing the second while with an if should do
the trick:

```
while (start < inptr && *start == '>') {

  start++;

  /* skip next space */
  if (start < inptr && *start == ' ') start++;
}

```

By the way, I plan to include this patch into https://github.com/gauteh/gmime/tree/blockquote_html for GMime.

Regards, Gaute

src/utils/gmime/gmime-filter-html-bq.c
```diff
@@ -219,9 +219,8 @@ citation_cut (char *in, const char *inend)
 	while (inptr < inend && *inptr != '\n') {
 		if (*inptr == ' ')
 			inptr++;
-		if (inptr >= inend || *inptr != '>') {
+		if (inptr >= inend || *inptr != '>')
```
gauteh: If you make this change, then inptr++ should be on the same line. (see heartbleed bug)

mxmehl:
What's the current status of this PR? With the conflicting files I'm averse to test this with my productive astroid installation, but I'd love to have this cosmetic bug resolved :) 

Can I help somehow to get this done?

gauteh:
mxmehl writes on desember 13, 2016 23:28:
> What's the current status of this PR? With the conflicting files I'm averse to test this with my productive astroid installation, but I'd love to have this cosmetic bug resolved :) 
> 
> Can I help somehow to get this done?

Shouldn't take too much effort to get this one ready, not sure why it
conficts - maybe the mid changes came afterwards (they are unrelated
anyway), but would probably mess up the patch a little bit.

The patch still needs a bit of coding, if you wish please take a look at
it.

mxmehl:
Am 14.12.2016 um 00:21 schrieb Gaute Hope:
> The patch still needs a bit of coding, if you wish please take a look at
> it.

Sorry, that's out of my league (like almost everything in astroid
unfortunately).

gauteh:
mxmehl writes on desember 14, 2016 1:27:
> Am 14.12.2016 um 00:21 schrieb Gaute Hope:
>> The patch still needs a bit of coding, if you wish please take a look at
>> it.
> 
> Sorry, that's out of my league (like almost everything in astroid
> unfortunately).

I'll see if I can find the time to fix the last issues. Then testing
would be very useful :)

gauteh:
Please test #247.

PR #157, opened 2016-06-13
Plugins: libpeas plugins

gauteh:
Check out [astroid-plugins/official-examples](https://github.com/astroidmail/astroid-plugins), for instance: `ThreadViewExamplePlugin/tagformat.py` and the method `format_tags`. Put this folder in `.config/astroid/plugins/` and run astroid through the wrapper script: `examples/astroid` (to make sure `GI_TYPELIB_PATH` is set when you do not install astroid).

Corresponding issue: #156.

Documentation:

## General

The environment variable `ASTROID_CONFIG` will be set by astroid, it points to the configuration file usually located in: `~/.config/astroid/config`.

##### Objects
`thread_view`: [GtkBox](https://developer.gnome.org/gtk3/stable/GtkBox.html) where the main TreeView is a child.
`web_view`: A [WebKitWebView](http://webkitgtk.org/reference/webkitgtk/stable/webkitgtk-webkitwebview.html)

##### Functions
`string uri = get_avatar_uri (string email, GMime.Message message)`
> make sure the uri (or the first part of it) is returned in `get_allowed_uris()`

`[list of strings] allowed_uris = get_allowed_uris ()`

`string html_markup = format_tags (string background_color, [list of strings] tags, bool selected)`
> See documentation for `ThreadIndex::format_tags plugin`.

##### Objects
`thread_index` : [GtkBox](https://developer.gnome.org/gtk3/stable/GtkBox.html) where the main TreeView is a child.

##### Functions
`string pango_markup = format_tags (string background_color, [list of strings] tags, bool selected)`
> This can be used to format tags just like you want: #139.
> background_color: (e.g. #ffffff) the color of the background the tag is rendered on (typically used for calculating luminosity when using alpha values)

PR #168, opened 2016-07-03
Crypto 2: Encryption, signing and proper visualization of encrypted messages

gauteh:


mxmehl:
Thank you very much! However, I have a problem testing this.

After checking out the crypto-2 branch, I executed `scons` but it failed with 

```
scons: Reading SConscript files ...
building version v0.5-282-g0db08743 (git)..
debug flag enabled: True
Checking for pkg-config... (cached) yes
Checking for gtkmm-3.0 >= 3.10... (cached) yes
Checking for glibmm-2.4... (cached) yes
Checking for gmime-2.6 >= 2.6.18... (cached) yes
Checking for webkitgtk-3.0... (cached) yes
Checking for C library libsass... (cached) no
Checking for C library libsass... (cached) yes
Checking for libpeas-1.0... no
libpeas-1.0 not found.
```

What do I have to install on an Arch Linux system? I already tried libpeas (version 1.18.0-1) from the extra repository, without success unfortunately.

gauteh:
Hm - libpeas should be enough. Maybe delete the `scons` cache: .scons*

gauteh:
Note that you will pull in the plugins branch ( #157), but that should work fine.

mxmehl:
Thanks for your quick reply!

> Hm - libpeas should be enough. Maybe delete the scons cache: .scons*

No success.

> Note that you will pull in the plugins branch ( #157), but that should work fine.

Can you elaborate please? I cloned the astroid repo, checkout crypto-2 branch and tried `scons`. Anything else that I would have to do beforehand?

gauteh:
mxmehl writes on juli 3, 2016 12:30:
> Thanks for your quick reply!
> 
>> Hm - libpeas should be enough. Maybe delete the scons cache: .scons*
> 
> No success.

What is the output of:

$ pkg-config --libs --cflags libpeas-1.0

?


>> Note that you will pull in the plugins branch ( #157), but that should work fine.
> 
> Can you elaborate please? I cloned the astroid repo, checkout crypto-2 branch and tried `scons`. Anything else that I would have to do beforehand?

No you should be fine, the crypto-2 branch is based on the plugins
branch, but you don't need to worry about that.

mxmehl:
Ah, your hint solved it. I had to install `gobject-introspection` as well, only `gobject-introspection-runtime` was installed. The output was:

```
Package gobject-introspection-1.0 was not found in the pkg-config search path.
Perhaps you should add the directory containing `gobject-introspection-1.0.pc'
to the PKG_CONFIG_PATH environment variable
Package 'gobject-introspection-1.0', required by 'libpeas-1.0', not found
```

Encryption seems to work now! Thanks a lot :)

gauteh:
mxmehl writes on juli 3, 2016 14:16:
> Ah, your hint solved it. I had to install `gobject-introspection` as well, only `gobject-introspection-runtime` was installed. The output was:
>
> ```
> Package gobject-introspection-1.0 was not found in the pkg-config search path.
> Perhaps you should add the directory containing `gobject-introspection-1.0.pc'
> to the PKG_CONFIG_PATH environment variable
> Package 'gobject-introspection-1.0', required by 'libpeas-1.0', not found
> ```

Ah, of course - I moved the gobject-introspection up first, then it
should be obvious.

>
> Encryption seems to work now! Thanks a lot :)

Great - I've added more signature display and stuff, might need more
visualization - especially of bad signatures.

Issues:
- Encrypted multi-parts inside encrypted parts doesn't work as the
  crypto object would be overwritten? It would at least not be obvious
  that the message has several layers of encryption (only the last one
  would be shown).

- Attachments might appear to be encrypted, but may be encrypted with
  other multi-part than main message (and is not necessarily valid or
  from the same person).

- Nothing else than gpg/pgp works (no inline).

- It is a pretty ugly yellow now.

- Failure in encryption is not handled very gracefully.

mxmehl:
In general, de- and encryption and signing works fine (which is awesome, it gets more and more usable for me!)

Some notes on my first tests:

- When encrypting a mail, it seems to add a `=` after each mail (I am using a signature). When only signing it, this doesn't happen.
- You wrote that one can see the status of the encryption somewhere (with a yellow box or something). Do I have to change my pre-encryption config somehow to see it? Right now I only see the signed and encrypted tags.
- Can one somehow configure the behaviour of the encryption? It would be awesome to have at least some basic rules like "if mail to me is encrypted, encrypt the reply" or "always sign messages". More advanced settings could be "do not sign every message but all those matching a regex (like .*@organization.com)"

Thanks again! I owe you some beers/coffees/teas ;)

gauteh:
mxmehl writes on juli 4, 2016 20:08:
> In general, de- and encryption and signing works fine (which is awesome, it gets more and more usable for me!)

Great!

 
> Some notes on my first tests:
> 
> - When encrypting a mail, it seems to add a `=` after each mail (I am using a signature). When only signing it, this doesn't happen.

I thought I got rid of that earlier.. somehow that returned.

> - You wrote that one can see the status of the encryption somewhere (with a yellow box or something). Do I have to change my pre-encryption config somehow to see it? Right now I only see the signed and encrypted tags.

Are you using the latest version of the crypto-2 branch? I pushed some
changes that displays those things (very basic atm).

> - Can one somehow configure the behaviour of the encryption? It would be awesome to have at least some basic rules like "if mail to me is encrypted, encrypt the reply" or "always sign messages". More advanced settings could be "do not sign every message but all those matching a regex (like .*@organization.com)"

Yes, that would be great.. I think the regexp stuff should rather go in
a plugin. The rest should be part of the account config. I assume you
already put the keyid in the config right?

 
> Thanks again! I owe you some beers/coffees/teas ;)
 

Nice, :cheers:!

-gaute

mxmehl:
# Gaute Hope [04.07.2016 @ 20:36]:
> mxmehl writes on juli 4, 2016 20:08:
>> - You wrote that one can see the status of the encryption somewhere
>> (with a yellow box or something). Do I have to change my pre-encryption
>> config somehow to see it? Right now I only see the signed and encrypted
>> tags.
>
> Are you using the latest version of the crypto-2 branch? I pushed some
> changes that displays those things (very basic atm).

Yes, I do. In the title bar, it shows v5.0-290-gd746607c.

When building it another time, I noticed two error/warning messages:

```
compiling ==> src/compose_message.cc
src/compose_message.cc: In member function 'void
Astroid::ComposeMessage::finalize()':
src/compose_message.cc:243:12: warning: variable 'res' set but not used
[-Wunused-but-set-variable]
       bool res = false;
            ^~~
```

and

```
compiling ==> src/plugin/gir_main.c
linking Program ==> girmain
LD_LIBRARY_PATH=./ g-ir-scanner -o Astroid-0.1.gir --warn-all
--namespace=Astroid --include=GObject-2.0 --include=GMime-2.6
--nsversion=0.1 --pkg=gobject-introspection-1.0 --pkg=gmime-2.6 -Isrc/
--program=./girmain src/plugin/astroid_activatable.c
src/plugin/gir_main.c src/plugin/thread_index_activatable.c
src/plugin/thread_view_activatable.c src/plugin/astroid_activatable.h
src/plugin/thread_index_activatable.h src/plugin/thread_view_activatable.h
/bin/libtool: line 2460: printf: write error: Bad file descriptor
src/plugin/thread_view_activatable.h:45: Warning: Astroid:
astroid_threadview_activatable_get_allowed_uris: return value: Missing
(element-type) annotation
g-ir-compiler -o Astroid-0.1.typelib Astroid-0.1.gir
linking Program ==> astroid
```

Encrypted messages still look quite normal, no background colour, signature list or similar.

>> - Can one somehow configure the behaviour of the encryption? It would
>> be awesome to have at least some basic rules like "if mail to me is
>> encrypted, encrypt the reply" or "always sign messages". More advanced
>> settings could be "do not sign every message but all those matching a
>> regex (like .*@organization.com)"
>
> Yes, that would be great.. I think the regexp stuff should rather go in
> a plugin. The rest should be part of the account config. 

Cool, having the basic stuff in the account config would be cool ;)

> I assume you already put the keyid in the config right?

Yes, the accounts part looks like this:

```
"name": "Max Mehl",
"email": "max.mehl@...",
"gpgkey": "0x371E2E92",
"sendmail": "msmtp --read-envelope-from -a fsfe -t",
"default": "true",
"save_sent": "true",
"save_sent_to": "\/home\/max\/Mails\/FSFE\/Sent\/cur\/",
"additional_sent_tags": "",
"save_drafts_to": "\/home\/max\/Mails\/FSFE\/Drafts\/cur\/",
"signature_file": "signature.txt",
"signature_default_on": "true",
"signature_attach": "false"
```

gauteh:
That looks fine, the warning will be fixed as part of the full
implementation. Haven't quite figured out the GIR error yet..

Does the yellow box show Signed by: ... as in this screenshot? It only
works for encrypted and signed messages (or only signed messages).

Regards, Gaute

mxmehl:
> Does the yellow box show Signed by: ... as in this screenshot? It only works for encrypted and signed messages (or only signed messages).

Oh wow, that look nice (but yes well, we can talk about the yellow :P)! Unfortunately on my side it's just plain, as you can see on my screenshot. Maybe on the weekend I set up a plain profile and test it with that...

astroid_signed

gauteh:
mxmehl writes on juli 5, 2016 22:02:
> astroid_signed

Weird - could you paste the astroid-log of you opening that email? Also,
just to make sure, pull latest 'crypto-2' branch and re-compile. It
almost seems like you are not using the latest version.

Also - it only supports gpg-signed messages, smime is ignored.

mxmehl:
Aaah, thanks for giving me the hint with the log window. It seemed the files of the "ui" folder in my config directory overwrote it. I just deleted it and now it looks fine! Please excuse the caused confusion...

gauteh:
mxmehl writes on juli 5, 2016 16:48:
> src/plugin/thread_view_activatable.h:45: Warning: Astroid:
> astroid_threadview_activatable_get_allowed_uris: return value: Missing
> (element-type) annotation

Fixed in 74777e12387948fa2d0fba0d542af39d6fa056fd.

mxmehl:
I just added a PR to change the look of the encrypted/signed containers around messages and attachments a bit. I hope I did this correctly (I'm quite unused to git workflows, sorry).

I have a few other suggestion which I cannot implement on my own:

- It seems the importance of signed is higher than encrypted when it comes to which colour the message container has. I suggest making it grey (or green as in my PR) even if it's encrypted *and* signed. Currently this results in a blue (orange) box around the message.
- Make the look of the encrypted_info (the yellow box) depending on the status of the encryption. For example, if the message is just signed, keep it yellow. If it is successfully encrypted, make it green. And if there's an error, make it redish. 
- As already written in IRC, astroid doesn't give a warning if an encrypted mail is written to a recipient for which the public key is unknown. This in kinda bad because I have no clue which persons' pubkeys I have ;)
- Even if they're bad, it would be nice to have Inline GPG messages supported, at least partly.

gauteh:
mxmehl writes on juli 7, 2016 11:21:
> I just added a PR to change the look of the encrypted/signed containers around messages and attachments a bit. I hope I did this correctly (I'm quite unused to git workflows, sorry).

Looks good!

> 
> I have a few other suggestion which I cannot implement on my own:
> 
> - It seems the importance of signed is higher than encrypted when it comes to which colour the message container has. I suggest making it grey (or green as in my PR) even if it's encrypted *and* signed. Currently this results in a blue (orange) box around the message.

You should be able to specify a CSS class which has both, I think that
should take precedence.

I added two CSS classes: .decrypt_failed and .verify_failed which you
can use to color failed decrypt and signature verifications differently
(e.g. some kind of red).

> - Make the look of the encrypted_info (the yellow box) depending on the status of the encryption. For example, if the message is just signed, keep it yellow. If it is successfully encrypted, make it green. And if there's an error, make it redish. 

^^^
see above.

If you haven't discovered it yourself already you can right click in the
thread-view and inspect elements, useful when working on these things.

> - As already written in IRC, astroid doesn't give a warning if an encrypted mail is written to a recipient for which the public key is unknown. This in kinda bad because I have no clue which persons' pubkeys I have ;)

Yes.

> - Even if they're bad, it would be nice to have Inline GPG messages supported, at least partly.

Decryption, yes. I agree.

mxmehl:
It seems the encryption is not as flawless as I thought initially. There were two people now who weren't able to decrypt my encrypted mails to them although I have their public keys. Sending them in Thunderbird with Enigmail worked however.

I am no GPG expert but it seems astroid is just encrypting the message with one's own key, not the public key of the recipient. At least that seems to be the striking difference between the mails that worked (only sent with TB so far) and those which didn't (astroid).

mxmehl:
> I am fairly sure decryption works as it should though?

Yes, flawlessly for GPG/MIME until now.

> I am working on showing some more information on who the message is encrypted too, that should make it easier to see if things are working.

Awesome, looking forward to that! And I hope that the encryption to recipients with known public keys will work too soon :)

Thanks again for your work!

gauteh:
mxmehl writes on juli 12, 2016 14:49:
>> I am fairly sure decryption works as it should though?
> 
> Yes, flawlessly for GPG/MIME until now.
> 
>> I am working on showing some more information on who the message is encrypted too, that should make it easier to see if things are working.
> 
> Awesome, looking forward to that! And I hope that the encryption to recipients with known public keys will work too soon :)
> 

There was an error in how the recipient list was generated, please try
again. Make sure that the receiving key is locally signed: gpg --lsign
before you can use it for encryption.

No luck yet in displaying recipient info, but that may be a gpg context
thing.

gauteh:
2016-07-13-130522_927x800_scrot

gauteh:
Note to self: If Bcc'ing it would be possible to get the recipients unless the keyid is anonymized.

emdete:
shouldn't BCCs encrypted and sent completly separate? then the ID would not reveal the adress...

gauteh:
- Decryption and verification errors are now shown.
- Encryption errors are shown.
- Always trust is on, --batch issues are reported upstream.
- Colors are configurable and more consistent.
- BCC is not sent separately, there should be a warning now - but you
  can do this by sending the email separately manually. I think that is
  how it is going to be now.
- BCC should shadow key ids, you can set this in you gnupg config.
- You do not need to lsign keys, but you do of course need to have the
  pub key of anything you want to verify or encrypt to.

mxmehl:
Wow, that's plain awesome, thank you very much!

The information boxes now show detailled information, that's great. Just one minor bug: In case of an error (no-pub-key in my example), the default colour for the box is still green. Seems to be a tiny logic error.

In this case there was also no border-left for the encrypted parts of the message. Is that intended? (not that I really care about it in case of an error)

20160714_auswahl_01

Personally, I changed the border-left to 2px dashed because it was a bit too dominant to me (and I wanted to make a distinction between the message-quote border-left) but that's a matter of taste :)

Now the only first-world-problem-wish I'd have is the feature to download pub keys from inside astroid in case on is unknown.

gauteh:
mxmehl writes on juli 14, 2016 12:19:
> Wow, that's plain awesome, thank you very much!
> 
> The information boxes now show detailled information, that's great. Just one minor bug: In case of an error (no-pub-key in my example), the default colour for the box is still green. Seems to be a tiny logic error.
> 
> In this case there was also no border-left for the encrypted parts of the message. Is that intended? (not that I really care about it in case of an error)

That is a verify_failed error, so the decryption still works - but the
signature failed. The box should not be green, but I've had encryption
status take precedence so far. The left-border should definetely not go
away (this logic is handled in the CSS).

> 
> Personally, I changed the border-left to 2px dashed because it was a bit too dominant to me (and I wanted to make a distinction between the message-quote border-left) but that's a matter of taste :)

Yes, if you find good colors / setup for this, please share. Did you
notice that you can change the color with SCSS variables?

> 
> Now the only first-world-problem-wish I'd have is the feature to download pub keys from inside astroid in case on is unknown. 

Try setting `crypto.gpg.auto_key_retrieve = true` in the config.

mxmehl:
> That is a verify_failed error, so the decryption still works - but the signature failed. The box should not be green, but I've had encryption status take precedence so far. 

Ah sure, that makes sense. I'm unsure which precendence would be the best... Enigmail solves this by being yellowish or grey AFAIR, but they don't really distinguish between signed and encrypted which is also kinda bad.

> Yes, if you find good colors / setup for this, please share. Did you notice that you can change the color with SCSS variables?

Yes, that's great! The default colors are very fine for me however.

>> Now the only first-world-problem-wish I'd have is the feature to download pub keys from inside astroid in case on is unknown.
> Try setting `crypto.gpg.auto_key_retrieve = true` in the config.

I can do, thanks. We talked about that issue in IRC yesterday but currently I cannot access the logs. Will I be informed about the downloaded key? And does it have performance issues as emdete noted?

gauteh:
mxmehl writes on juli 14, 2016 12:59:
>>> Now the only first-world-problem-wish I'd have is the feature to download pub keys from inside astroid in case on is unknown.
>> Try setting `crypto.gpg.auto_key_retrieve = true` in the config.
> 
> I can do, thanks. We talked about that issue in IRC yesterday but currently I cannot access the logs. 
> Will I be informed about the downloaded key? 

No.

> And does it have performance issues as emdete noted?

The GUI will block for the key to be downloaded.

mxmehl:
> Try setting `crypto.gpg.auto_key_retrieve = true` in the config.

I set this, deleted your public key and tried to view the message again. Astroid doesn't seem to start a download of the unknown pub-key. [Log here](https://p.mehl.mx/?53c1c2ccc8750065#EfpvPaZMOVQaGZMPnO/0gp56pyxdYeXSKeMXiy7dTnw=)

In gpg.conf I set a keyserver and my default-key. Do I have to add anything else there?

Thanks!

gauteh:
@emdete: did you get auto_key_retrive to work?

gauteh:
By the way, you have to restart astroid when you change the config.

gauteh:
Does it work for encryption?

gauteh:
@mxmehl: you also need to have keyserver-option auto-key-retrieve in .gnupg/gpg.conf set: https://github.com/astroidmail/astroid/wiki/Signing,-Encrypting-and-Decrypting

It will only work for signature verification though, not for encryption - there you have to fetch the key manually.

mxmehl:
@gauteh Ah, the config option for gpg.conf was missing, so now it works!

Interestingly, if the setting in gpg.conf is set, the keys are still being retrieved even if the astroid's auto_key_retrieve is set to false. I guess this is because gpg configs take precedence?

By the way, I'm looking forward to the inline PGP plan, sounds exciting :)

gauteh:
mxmehl writes on juli 15, 2016 16:49:
> @gauteh Ah, the config option for gpg.conf was missing, so now it works!
> 
> Interestingly, if the setting in gpg.conf is set, the keys are still being retrieved even if the astroid's auto_key_retrieve is set to false. I guess this is because gpg configs take precedence?

Yes, but I think gmime is supposed to disable it if it is set to false.
Does not seem to work.

 
> By the way, I'm looking forward to the inline PGP plan, sounds exciting :)

Ah, I pushed that :)

mxmehl:
Small bug: In case of an encrypted mail, astroid still attaches a `=` to the signature.

Apart from that, the PGP/MIME de- and encryption works very well up to now! :)

gauteh:
mxmehl writes on juli 17, 2016 12:42:
> Apart from that, the PGP/MIME de- and encryption works very well up to now! :)

Great!

> Small bug: In case of an encrypted mail, astroid still attaches a `=` to the signature.
 
Could you send me an example? Does your signature-file have a newline at
the end? Is the signature inlined or attached?

Is this only with signatures or does it happen when there is no
signature too?

mxmehl:
> Could you send me an example? 
I just sent you a batch of emails.

> Does your signature-file have a newline at the end? 
The signature has a newline at the end (most graphical editors automatically add one after saving a file)

> Is the signature inlined or attached? 
The signature is inline, not attached.

> Is this only with signatures or does it happen when there is no signature too?
As you can see, it is also w/o a signature, even if I took care that there's no empty line attached (at least in gvim).

For me as a outsider, it seems that empty lines at the very end get replaced by `=` somehow if they are encrypted (doesn't matter if signed or not).

gauteh:
mxmehl writes on juli 18, 2016 17:13:
>> Could you send me an example?
> I just sent you a batch of emails.
>
>> Does your signature-file have a newline at the end?
> The signature has a newline at the end (most graphical editors automatically add one after saving a file)
>
>> Is the signature inlined or attached?
> The signature is inline, not attached.
>
>> Is this only with signatures or does it happen when there is no signature too?
> As you can see, it is also w/o a signature, even if I took care that there's no empty line attached (at least in gvim).
>
> For me as a outsider, it seems that empty lines at the very end get replaced by `=` somehow if they are encrypted (doesn't matter if signed or not).

So, after reading more about quoted-printable it seems that the '='
should be added as soft line-breaks and some other cases [0] (notably if
lines are longer than 76 chars). It seems GMime often justs adds it for
good measure [1].

I cannot see the extra '=' in the decoded text, only in the encoded text
(using 'V' on the draft or the sent message) - screenshot attached. What
version of GMime do you have?

Make sure you use: git pull --force on the crypto-2 branch, I re-based
it on master.

[0] https://en.wikipedia.org/wiki/Quoted-printable
[1] https://www.mail-archive.com/gmime-devel-list@gnome.org/msg00086.html

gauteh:
2016-07-18-180316_1200x1526_scrot

mxmehl:
>  I cannot see the extra '=' in the decoded text, only in the encoded text (using 'V' on the draft or the sent message) - screenshot attached. 

Same here, I cannot see the `=` in astroid, but for example in Thunderbird (which many of my contact have). 

> What version of GMime do you have?

2.6.20-2 (arch linux, from extra rep)

Can you maybe send me an encrypted test mail with inline signature like I did? I'm curious whether I can see the same effect I have with my mails.

gauteh:
mxmehl writes on juli 18, 2016 18:36:
>>  I cannot see the extra '=' in the decoded text, only in the encoded text (using 'V' on the draft or the sent message) - screenshot attached. 
> 
> Same here, I cannot see the `=` in astroid, but for example in Thunderbird (which many of my contact have). 

So if this only happens with encrypted email it might be an issue with
Enigmail decoding (from quoted-printable) the decrypted message. If the
receivers only see the last '=', but no other '=' (e.g. from any
umlauts) it seems that it is an issue with processing, not with
identifying the actual content.

>> What version of GMime do you have?
> 
> 2.6.20-2 (arch linux, from extra rep)

Ok - so same as me. Anyway, the gmime e-mails I referenced are from 2010.

> Can you maybe send me an encrypted test mail with inline signature like I did? I'm curious whether I can see the same effect I have with my mails.
> 

It is done. Do you have Thunderbird set up to inspect it?

mxmehl:
> It is done. Do you have Thunderbird set up to inspect it?

Yes, please see the scrot to see how it looks like. I have never experienced that before, not only with my mails but also with those I received (e.g. sent by mutt)

20160718_auswahl_01

gauteh:
If you compose a message with your inline signature and choose 'V' when
viewing it, a temporary copy will be written to /tmp/astroid-compose-XXX
- if you inspect the copy you should see the encoded content with the
Content-Transfer-Encoding line.

If you encrypt it, view it with 'V' and inspect the temprorary copy in
/tmp through 'gpg -d' you should still be able to see the
Content-Transfer-Encoding line. Is there any difference between the two
copies? It doesn't seem to be on my part.

(the temporary copy is deleted when you close the raw view, inspect the
actual file to be able to decrypt it with gpg, since the raw view will
show the encrypted message).

You need latest crypto-2 for this to work.

gauteh:
mxmehl writes on juli 18, 2016 18:46:
>> It is done. Do you have Thunderbird set up to inspect it?
> 
> Yes, please see the scrot to see how it looks like. I have never experienced that before, not only with my mails but also with those I received (e.g. sent by mutt)

```
hello there.
--  =20
<Esc>:Wq!:wq:q!
 =20
=
```

As you can see in the raw view there are lots of quoted-printable
encoded stuff there already, which enigmal + thunderbird seems to handle
- it is weird that it does not handle the last '='. Does the PGP signature verify correctly in Thunderbird? That means that it gets the content right at least.

mxmehl:
> As you can see in the raw view there are lots of quoted-printable encoded stuff there already, which enigmal + thunderbird seems to handle - it is weird that it does not handle the last '='.

Strange indeed. I don't know how to solve that. I just sent you a reply with TB. Maybe you can tell the difference from that.

> Does the PGP signature verify correctly in Thunderbird? That means that it gets the content right at least.

Yes, this seems to be the case up to now and with several test and from various senders/clients.

gauteh:
Gaute Hope writes on juli 18, 2016 18:51:
> ```
> hello there.
> --  =20
> <Esc>:Wq!:wq:q!
>  =20
> =
> ```

Works here: http://www.convertstring.com/EncodeDecode/QuotedPrintableDecode

gauteh:
Enigmail issue: https://sourceforge.net/p/enigmail/bugs/628/

So to sum up:
- sign + encrypt: extra '='
- sign: no extra
- encrypt: extra '='
- normal: no extra

gauteh:
Gaute Hope writes on juli 18, 2016 23:16:
> Enigmail issue: https://sourceforge.net/p/enigmail/bugs/628/

'='-issue is now fixed in Enigmail: https://sourceforge.net/p/enigmail/bugs/628/#1a17

mxmehl:
> '='-issue is now fixed in Enigmail: https://sourceforge.net/p/enigmail/bugs/628/#1a17

Excellent, thank you very much for reporting the bug. I am looking forward to the fix coming to effect in the next Enigmail release.

gauteh:
Merged, Inline PGP is tricky - will have to wait, but could also be done in a plugin.

PR #169, opened 2016-07-07
crypto-2: ui: colour changes and padding to crypto containers

mxmehl:
use orange for signed, green for encrypted
add padding between border and text to containers
make border thinner (1px instead of 2px)

### DIFF
--- a/ui/thread-view.scss
+++ b/ui/thread-view.scss
@@ -634,11 +634,11 @@ body.nohide .email .compressed_note > span {
 }
 
 .attachment .encrypted {
-  border: 1px dashed gray !important;
+  border: 1px dashed green !important;
 }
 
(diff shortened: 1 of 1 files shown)


gauteh:
mxmehl writes on juli 7, 2016 11:12:
> add padding between border and text to containers

One comment on this; does this break the 80 char width that emails are
usually formatted in?

mxmehl:
> One comment on this; does this break the 80 char width that emails are usually formatted in?

No, doesn't seem so. I looked at a few dozen signed/encrypted emails and this doesn't have a negative effect.

gauteh:
Thanks, merged! Could maybe use the same background color for the
encryption box (get rid of the yellow).

Issue #128, opened 2016-05-10
Print css : font size


hugoroy:
Try to print an email. See the font size is too big.

gauteh:
Hugo Roy writes on mai 10, 2016 19:57:
> Try to print an email. See the font size is too big.

This hit me too a few times.. can probably be fixed in the css (as you
suggest).

hugoroy:
↪ mai 10, 2016 11:28 — Gaute Hope:
> Hugo Roy writes on mai 10, 2016 19:57:
>> Try to print an email. See the font size is too big.
> 
> This hit me too a few times.. can probably be fixed in the css (as you
> suggest). 

I tried adding a font-size but it did not affect it. I have to look
deeper :)

gauteh:
Fixed in #178.

PR #178, opened 2016-07-23
Print size 11px

hugoroy:


gauteh:
Hugo Roy writes on juli 23, 2016 18:40:
> -- Commit Summary --
> 
>   * print view: not too big
>   * fix print font size issue #128

Thanks, merged!

-gaute

PR #181, opened 2016-07-25
Open drop-down terminal with '|'

gauteh:
Open a drop-down terminal with '|', useful in combination with 'Y' and patches.

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -236,6 +236,10 @@ if not conf.CheckPKG('webkitgtk-3.0'):
   print "webkitgtk not found."
   Exit (1)
 
+if not conf.CheckPKG ('vte-2.91'):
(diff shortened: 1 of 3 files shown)


mxmehl:
Works for me and nice to have.

Is it possible to swap the focus between terminal and the rest of the window? 
And is it correct that the terminal will only close by typing "exit" in the terminal itself, or pressing C-d?

gauteh:
thanks, Currently not. At the moment I have more or less decided not to include
this patch since it is just as easy and useful to just open a terminal
with a hotkey and use that.

Check out 'y', 'Y' and 'C-y' keybindings for copying message contents.

mxmehl writes on desember 23, 2016 16:38:

PR #187, opened 2016-08-09
cm, em: make message-id customizable, and user-agent optional

gauteh:
Inspired by 1470699317-30598-1-git-send-email-dkg@fifthhorseman.net on
notmuch@:

Message-id is now configurable, so that it is possible to remove any
traces of local setup (local hostname) or obvious traces of the
user-agent ('astroid' in the message-id).

Inclusion of the User-Agent in new messages is also optional and
configurable.

References:
https://www.gnu.org/software/emacs/manual/html_node/gnus/FAQ-5_002d13.html

### DIFF
--- a/History.txt
+++ b/History.txt
@@ -8,7 +8,8 @@
 
 * Make id:Message-ID urls clickable.
 
-* Make Message-ID customizable, and User-Agent optional.
+* Make Message-ID customizable, and User-Agent optional. These may also
+  be customized with a plugin.
 
 == v0.6 / 2016-07-23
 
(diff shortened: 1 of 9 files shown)


emdete:
very nice idea! maybe it could go even further by the ability to set the useragent from config or (even better) have an API to write an function in a plugin to determine useragent and message id (that way the agent and id-algo. could change with every mail you write)?

mxmehl:
Very good approach with the message-id! Furthermore I agree to @emdete's proposal with customizing the user agent as well.

gauteh:
@emdete has implemented a plugin randomizing these parameters here: https://github.com/astroidmail/astroid-plugin-seccomp

PR #193, opened 2016-08-14
Move to system's default sans-serif font

hugoroy:
I didn't realise that my personal font preferences were still in the default astroid message style. This removes references to custom fonts and instead use the system's default fonts. This also moves to a sans serif font rather than a monospace font by default.


gauteh:
thanks, merged!

Issue #180, opened 2016-07-24
More compact thread view tree


mxmehl:
With large threads (>30 messages) I tend to loose overview in astroid unfortunately. When opening an unread thread in Thread Index, I can hardly tell which emails in this thread are new and it's a lot of scrolling until I can be sure that I've seen every new mail. So I propose following solutions:

1. Make the thread view more compact (by config setting or shortcut). For a rough overview it would be sufficient to only have one line per mail (Subject, From, Date) and a marking each new/unread mail by making it bold (similar to how mutt or Thunderbird looks like).

2. Have a shortcut in thread view to fade out all already read emails and only show those which are new. Then one can be sure that every email in the current tab is worth to be read. However, that wouldn't be the best solutions since it's complicated to open older, referenced mails when needed.

Maybe related, maybe another issue: In Thread Index it would be nice to see the number of unread messages in a thread, not only the total number. Could also help to know how many unread messages one has to find ;)

hugoroy:
I'm not sure I get what you mean in 1. 

Are you saying we should make this more compact?
![capture d ecran de 2016-08-07 00-24-57](https://cloud.githubusercontent.com/assets/1629771/17459383/6538c8a2-5c35-11e6-9750-e78d8465c783.png)

gauteh:
mxmehl writes on juli 24, 2016 20:36:
> 1. Make the thread view more compact (by config setting or shortcut). For a rough overview it would be sufficient to only have one line per mail (Subject, From, Date) and a marking each new/unread mail by making it bold (similar to how mutt or Thunderbird looks like).
 
Related: you can use C-F to toggle flat or tree view of messages.
This helps me. 

> 2. Have a shortcut in thread view to fade out all already read emails and only show those which are new. Then one can be sure that every email in the current tab is worth to be read. However, that wouldn't be the best solutions since it's complicated to open older, referenced mails when needed.


Related: It would be nicer to make messages unread when they are focused
for some short time-delay, not when the whole thread is opened. That
would make it easier to keep track of what is not read (issue #160)

mxmehl:
Sorry for the late reply:

# @hugoroy

> Are you saying we should make this more compact?

Yes, that's what I meant. Currently we have around 4 lines for each mail. Reducing that to one or two lines with a slick design would help to improve overview in large threads. 

# @gauteh 

> Related: you can use C-F to toggle flat or tree view of messages. This helps me.

[...]

Issue #160, opened 2016-06-14
thread-view: mark messages unread on focus


gauteh:
thread-view should mark messages unread when focused (for e.g. 2 seconds, though I prefer them to be instantly marked unread), not just unmark the whole thread on load.

gauteh:
Implemented.

PR #195, opened 2016-08-15
Tv unread timeout

gauteh:
#160, #180.


mxmehl:
Using this since 2 days, no problems, nice feature ;)

PR #201, opened 2016-08-24
Improve poll.sh example

SuprDewd:
This is an attempt at improving the poll.sh example. In particular:
- Use `/usr/bin/env` bash instead of `/bin/bash`, as that is more portable
- Use `set -e` instead of `|| exit $?` after each command
- Add comments and examples

A few notes and questions before this is ready:
- Are there more examples we want to include? Whether we include them here or not, it would be nice to know about them. At the very least someone should add them to the Wiki.
- Am I using `notifymuch` correctly? I didn't manage to install it, and so haven't tested that piece of code.
- @gauteh had a mechanism for tagging only incoming mail. Specifically, there was a tag 'new' on incoming mail that he could target in his tagging script, which he then removed at the end of the script. As far as I can see this isn't standard [[1]](https://notmuchmail.org/pipermail/notmuch/2009/000767.html), and I saw some arguments against doing this. Also wonder if it makes the example unnecessarily complex. Maybe you could elaborate on where/how the 'new' tag is added?
- Does someone have better examples for the automatic tagging script?

### DIFF
--- a/examples/poll.sh
+++ b/examples/poll.sh
@@ -1,8 +1,42 @@
-#! /bin/bash
+#!/usr/bin/env bash
+#
+# poll.sh - An example poll script for astroid.
+#
+# Intended to be run by astroid. See the following link for more information:
+# https://github.com/astroidmail/astroid/wiki/Polling
+#
 
-# poll script for astroid
+# Exit as soon as one of the commands fail.
+set -e
 
-offlineimap || exit $?
 
-notmuch new || exit $?
+# Fetch new mail.
+offlineimap
+
+# Import new mail into the notmuch database.
+notmuch new
+
+
+# Here follow some examples of processing the new mail. The respective section
+# of code needs to be uncommented to be enabled. See also notmuch-hooks(5) for
+# an alternative place to put these commands.
+
+
+# Automatic tagging of new mail. See notmuch-tag(1) for details.
+
+#notmuch tag --batch <<EOF
+#
+# # Tag urgent mail
+# +urgent subject:URGENT
+#
+# # Tag all mail from GitHub as "work"
+# +work from:github
+#
+#EOF
+
+
+# Notify user of new mail. See the following link for more information:
+# https://github.com/astroidmail/astroid/wiki/Desktop-notification
+
+#notifymuch
 
--- a/examples/poll.sh
+++ b/examples/poll.sh
@@ -5,10 +5,12 @@
 # Intended to be run by astroid. See the following link for more information:
 # https://github.com/astroidmail/astroid/wiki/Polling
 #
+# In particular, in order for this script to be run by astroid, it needs to be
+# located at ~/.config/astroid/poll.sh
+#
 
 # Exit as soon as one of the commands fail.
 set -e
-
 
 # Fetch new mail.
 offlineimap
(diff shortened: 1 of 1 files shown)


gauteh:
Bjarki Ágúst Guðmundsson writes on august 24, 2016 20:15:
> This is an attempt at improving the poll.sh example. In particular:
> - Use `/usr/bin/env` bash instead of `/bin/bash`, as that is more portable
> - Use `set -e` instead of `|| exit $?` after each command
> - Add comments and examples

^^ Good.
 
> A few notes and questions before this is ready:
> - Are there more examples we want to include? Whether we include them here or not, it would be nice to know about them. At the very least someone should add them to the Wiki.

Maybe link to notmuch-wiki page referenced below. Other stuff that I
use:

I tag email arriving at different accounts with a tag using the to: query.
I tag mailing lists using the lists: query.
I tag email from certain people.

Might also be worth including the stuff about excluding threads:
https://notmuchmail.org/excluding/

> - Am I using `notifymuch` correctly? I didn't manage to install it, and so haven't tested that piece of code.

Yes, maybe make the notifymuch command work if the file is found?
Something like `which notifymuch && notifymuch`?

> - @gauteh had a mechanism for tagging only incoming mail. Specifically, there was a tag 'new' on incoming mail that he could target in his tagging script, which he then removed at the end of the script. As far as I can see this isn't standard [[1]](https://notmuchmail.org/pipermail/notmuch/2009/000767.html), and I saw some arguments against doing this. Also wonder if it makes the example unnecessarily complex. But could you at least elaborate on where/how the 'new' tag is added?

This is related to the new tag approach described here: https://notmuchmail.org/initial_tagging/, I think this page should be referenced somewhere anyway.

I think this (or equivalent) should be the suggested approach, since
manually changing a tag later (say one query mis-tags something) would
be undone each time the script is run.

Maybe we should keep both a super-simple poll.sh script and one extended one?

gauteh:
Gaute Hope writes on august 25, 2016 0:14:
> Bjarki Ágúst Guðmundsson writes on august 24, 2016 20:15:
>> A few notes and questions before this is ready:
>> - Are there more examples we want to include? Whether we include them here or not, it would be nice to know about them. At the very least someone should add them to the Wiki.

I also have this check for an internet connection first:

```
# check if we have a connection
if ! ping -w 1 -W 1 -c 1 mail.google.com; then
  fail "there is no internet connection."
fi
```

And, since I have my mail mounted as an encfs:

```
# check if .mail is mounted
if [ ! -d ~/.mail/gaute.vetsj.com ]; then
  fail ".mail does not seem to be mounted."
fi
```

SuprDewd:
I went with your suggestion and kept the poll.sh example as simple as it used to be. So currently this PR just adds some comments to poll.sh, as well as makes it more portable (the old version doesn't work on NixOS).

I then extracted all the extra functionality into a Wiki page that poll.sh references: [Processing mail](https://github.com/astroidmail/astroid/wiki/Processing-mail).

What do you think?

gauteh:
Great! This looks good, and also a very nice wiki page! The only comment on that one is that the `notmuch tag` command example should work on only `tag:new` messages, could you update it to do that?

gauteh:
Merged.

SuprDewd:
> The only comment on that one is that the notmuch tag command example should work on only tag:new messages, could you update it to do that?

Updated. Can you have a look? I haven't set this up myself, so this is guesswork to some extent.

gauteh:
I dont use the --batch method, think it is easier if you test. Ill have a
look afterwards. Thanks.

søn. 12. feb. 2017 kl. 17.58 skrev Bjarki Ágúst Guðmundsson <
notifications@github.com>:

PR #202, opened 2016-08-27
Remove focus style and style attachments

hugoroy:
I would also suggest serious change to the UI of showing information regarding crypto.

### DIFF
--- a/ui/thread-view.scss
+++ b/ui/thread-view.scss
@@ -28,6 +28,14 @@
   $sent-collapsed: #f5f5f5 !global;
 }
 
+/* Colors associated with attachments: */
+@if not(global-variable-exists(attachment-color)) {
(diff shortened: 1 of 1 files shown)


gauteh:
It seems that long links are not wrapped, was this the behavior previously as well?
2016-08-28-160832_1200x1521_scrot

hugoroy:
↪ août 28, 2016 4:10 — Gaute Hope:
> It seems that long links are not wrapped, was this the behavior previously as well?
> 2016-08-28-160832_1200x1521_scrot
> 

I don't remember, but I do not see how these commits would have affected this.

The last commit should fix that :)

hugoroy:
@gauteh I think this is ready now :-)

gauteh:
Hugo Roy writes on september 1, 2016 20:32:
> @gauteh I think this is ready now :-)

Thanks, merged.

PR #207, opened 2016-08-31
change from () to <> for email display

emdete:
i consider <> as a convention / pattern to display / denote email addresses so i changed it in astroid accordingly. what do you think?


gauteh:
Thanks. I'll merge in a bit unless someone has any complaints.

Issue #212, opened 2016-09-07
Unable to add "BackSpace" as a keybinding


Radvendii:
There seems to be a hardcoded list of special keys (return, arrow keys, question mark, etc) that are allowed in keybindings, and have other keywords that are mapped to the GDK_KEY_whatever. First of all, is there a way to allow arbitrary GDK_KEY_whatever variables in the keybindings file? Alternatively, should someone (I can do this) add a comprehensive list? Also, this is documented nowhere. The wiki says to use devel/get_keys.py to get the list of default keybindings, but this generates a list with, for example,

`# searches.open=Key (GDK_KEY_Return) # Open query, default: Key (GDK_KEY_Return), Key (GDK_KEY_KP_Enter)`

which, if uncommented, doesn't run (since the proper command would be searches.open=Return
Another issue with the documentation is that it is totally unclear how to set multiple keys to the same action. The commented out lines don't even show this being done, but just say that it has been done.

I am happy to make some of these changes, I am just new to the community and wanted to know what peoples' feelings are about this and what this should be changed to.

gauteh:
Radvendii writes on september 7, 2016 2:47:
> There seems to be a hardcoded list of special keys (return, arrow keys, question mark, etc) that are allowed in keybindings, and have other keywords that are mapped to the GDK_KEY_whatever. First of all, is there a way to allow arbitrary GDK_KEY_whatever variables in the keybindings file? Alternatively, should someone (I can do this) add a comprehensive list? Also, this is documented nowhere. The wiki says to use devel/get_keys.py to get the list of default keybindings, but this generates a list with, for example,
> 
> `# searches.open=Key (GDK_KEY_Return) # Open query, default: Key (GDK_KEY_Return), Key (GDK_KEY_KP_Enter)`
> 
> which, if uncommented, doesn't run (since the proper command would be searches.open=Return
> Another issue with the documentation is that it is totally unclear how to set multiple keys to the same action. The commented out lines don't even show this being done, but just say that it has been done.
> 
> I am happy to make some of these changes, I am just new to the community and wanted to know what peoples' feelings are about this and what this should be changed to.

Hi,

It would be great to get a proper system for this, the hardcoded list
was something of a quick fix. There is the:

`gdk_keval_from_name ()` and `gdk_keyval_name ()` functions that can be
used for this: https://developer.gnome.org/gdk3/stable/gdk3-Keyboard-Handling.html#gdk-keyval-from-name.

[...]

PR #216, opened 2016-09-09
made keybindings general

Radvendii:
I changed the kludgy list of keynames_to_keyval for the general gdk function gdk_keyval_from_name and gdk_keyval_name.

For some reason I'm having trouble compiling master, but I've tested the changes with v0.6 and they work. Can someone else test with master? If the changes work, it may be possible to remove keynames_to_keyvals altogether, but I wasn't sure if other things were relying on it.
solves issue #212

### DIFF
--- a/src/modes/keybindings.cc
+++ b/src/modes/keybindings.cc
@@ -122,6 +122,10 @@ namespace Astroid {
             UstringUtils::trim (target);
 
             Key k (keyspec);
+            if (k.key == GDK_KEY_VoidSymbol) {
+              LOG (error) << "ky: user bindings: invalid key name: " << spec;
+              continue;
+            }
 
             LOG (debug) << "ky: run: " << name << "(" << k.str () << "): " << target;
 
@@ -553,7 +557,7 @@ namespace Astroid {
     } else {
       /* check if key is in map */
       UstringUtils::trim (k);
-      const char *k_str = k.c_str ();
+      const char * k_str = k.c_str ();
       guint kk = gdk_keyval_from_name (k_str);
 
       return kk;
@@ -674,11 +678,12 @@ namespace Astroid {
     if (isgraph (k)) {
       s += k;
     } else {
-      try {
-        ustring kk = gdk_keyval_name (key);
-        s += kk;
-      } catch (std::exception &ex) {
-        s += ustring::compose ("%1", key);
+      gchar * c = gdk_keyval_name (key);
+      if (c == NULL) {
+        LOG (error) << "invalid key: " << key << " for: " << name;
+        throw keyspec_error ("invalid key");
+      } else {
+        s += ustring (c);
       }
     }
 
--- a/src/modes/keybindings.cc
+++ b/src/modes/keybindings.cc
@@ -574,7 +574,8 @@ namespace Astroid {
     } else {
       /* check if key is in map */
       UstringUtils::trim (k);
-      guint kk = Keybindings::keynames_to_keyval.at (k);
+      const char *k_str = k.c_str ();
+      guint kk = gdk_keyval_from_name (k_str);
 
       return kk;
     }
@@ -695,7 +696,7 @@ namespace Astroid {
       s += k;
     } else {
       try {
-        ustring kk = Keybindings::keyval_to_keynames.at (key);
+        ustring kk = gdk_keyval_name (key);
         s += kk;
       } catch (std::exception &ex) {
         s += ustring::compose ("%1", key);
--- a/src/modes/keybindings.cc
+++ b/src/modes/keybindings.cc
@@ -19,32 +19,7 @@ namespace Astroid {
   std::vector<Key>  Keybindings::user_bindings;
   std::vector<std::pair<Key, ustring>> Keybindings::user_run_bindings;
 
-  std::map<guint, ustring> Keybindings::keyval_to_keynames = {
-    { GDK_KEY_Down,   "Down" },
-    { GDK_KEY_Up,     "Up" },
-    { GDK_KEY_Tab,    "Tab" },
-    { GDK_KEY_ISO_Left_Tab, "ISO_Left_Tab" },
-    { GDK_KEY_Home,   "Home" },
-    { GDK_KEY_End,    "End" },
-    { GDK_KEY_Return, "Return" },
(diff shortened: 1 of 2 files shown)


gauteh:
Thanks, what are the compile-problems?

Some comments:
* You need to handle errors in both these functions, what if the user
  inputs some random name?
* You can remove keynames_to_keyvals, but grep the source code to see if
  it is used anywhere. This inlcludes the reverse map that is set up in
  ::init ().


Regards, Gaute

Radvendii:
The compile problems probably have to do with the fact that I'm using nixos (almost surely) -- that tends to cause compile problems. My friend had to patch a few things to get v0.6 working, so I'm not surprised it doesn't work on master. When the next release comes out, we'll get it to compile, until then I don't know that it's worth it.

* Okay, How should I handle the errors? It seems like if they input some random string, gdk_keyval_from_name() will return GDK_KEY_VoidSymbol. Should I warn, or error if that happens? Isn't that the way to unset keybindings?
  The other function call I don't think needs error handling, the erroneous keybindings should have already been dealt with at that point right?
  Another interesting point is that there wasn't any error checking in these functions before. Which isn't necessarily to say that I shouldn't add it, I'm just wondering if there's a reason it was deemed unnecessary

* Good call. I made sure it wasn't used and removed it.

gauteh:
Radvendii writes on september 9, 2016 15:22:
> The compile problems probably have to do with the fact that I'm using nixos (almost surely) -- that tends to cause compile problems. My friend had to patch a few things to get v0.6 working, so I'm not surprised it doesn't work on master. When the next release comes out, we'll get it to compile, until then I don't know that it's worth it.

Ok, if these are patches that could go into the main repo without
interfering with 'standard' unix layout then I'd be happy to include
them.

> * Okay, How should I handle the errors? It seems like if they input some random string, gdk_keyval_from_name() will return GDK_KEY_VoidSymbol. Should I warn, or error if that happens? Isn't that the way to unset keybindings?

I think this should be an error, it does not need to crash astroid, but
it should stop adding that key binding.

>   The other function call I don't think needs error handling, the erroneous keybindings should have already been dealt with at that point right?

I guess not, maybe throw a logical_error or something?

>   Another interesting point is that there wasn't any error checking in these functions before. Which isn't necessarily to say that I shouldn't add it, I'm just wondering if there's a reason it was deemed unnecessary

Hm, yeah, you are right. We just need a good way to inform the user that
he input some garbage and then preferably drop the keybinding so that it
is not used. Maybe set an error flag on the Key, don't add it, and check
after line 149, 184, and any other places that constructor is called,
if the key could be parsed.

>
> * Good call. I made sure it wasn't used and removed it.

Nice, I think this is almost ready :+1:.

-gaute

Radvendii:
I think the patch could actually be safely added to the main project. I'll work on that pull request after this

Okay, It now logs an error and skips the keybinding, but doesn't kill the program.

I'm just wondering if the logical_error business is more code clutter than it's worth. I don't have a lot of experience maintaining code, but it strikes me that the worst thing that happens if the error isn't caught (and the error shouldn't even happen in the first place) is that there's a VoidSymbol keybinding generated by `devel/get_keys.py`. is there something I'm missing?

Anything I missed? Not sure what you wanted me to add after 149,184, but I think the error checking I did add might have covered it.

Thanks so much for your help : )

gauteh:
Radvendii writes on september 11, 2016 3:48:
> I think the patch could actually be safely added to the main project. I'll work on that pull request after this

Great!


> Okay, It now logs an error and skips the keybinding, but doesn't kill the program.

Great, have to make the same check after line 124.


> 
> I'm just wondering if the logical_error business is more code clutter than it's worth. I don't have a lot of experience maintaining code, but it strikes me that the worst thing that happens if the error isn't caught (and the error shouldn't even happen in the first place) is that there's a VoidSymbol keybinding generated by `devel/get_keys.py`. is there something I'm missing?

I mean something like this: 

```
diff --git a/src/modes/keybindings.cc b/src/modes/keybindings.cc
index fcd6eab..d2c752a 100644
--- a/src/modes/keybindings.cc
+++ b/src/modes/keybindings.cc
@@ -674,11 +674,13 @@ namespace Astroid {
     if (isgraph (k)) {
       s += k;
     } else {
-      try {
-        ustring kk = gdk_keyval_name (key);
-        s += kk;
-      } catch (std::exception &ex) {
-        s += ustring::compose ("%1", key);
+
+      gchar * c = gdk_keyval_name (key);
+      if (c == NULL) {
+        LOG (error) << "invalid key: " << key << " for: " << name;
+        throw keyspec_error ("invalid key");
+      } else {
+        s += ustring (c);
       }
     }
```

The exception stuff was left over from the keyval to keyname map. The
point is to catch programming errors on the first run. So that if I add
a new key, compile and start astroid, it will immediately crash rather
than giving me the impression that everything worked. Admittedly, it
only crashes if it logs the key, but I still think this is better :)
 

> Anything I missed? Not sure what you wanted me to add after 149,184, but I think the error checking I did add might have covered it.

I mixed up the lines, related to your check above.

> Thanks so much for your help : )

Thanks for the patch!

Radvendii:
Ah okay! that makes sense. Sorry I missed line 124, I've actually never programmed in C++ before, so I had no idea what that was doing.
Everything should be up to snuff.

gauteh:
Radvendii writes on september 11, 2016 16:27:
> Ah okay! that makes sense. Sorry I missed line 124, I've actually never programmed in C++ before, so I had no idea what that was doing.
> Everything should be up to snuff.

Impressive. I fixed up somme of the commit messages, and squashed some
of the commits before merging. Have a look at the git-log for master for
more details. 

Would be nice with some documentation about the available key bindings
now, also that they are case-sensitive.

Regards, Gaute

gauteh:
Hi, just got this bug report on AUR:


Hi,

getting this when compiling:

src/modes/keybindings.cc: In static member function 'static void
Astroid::Keybindings::init()':
src/modes/keybindings.cc:126:73: error: 'spec' was not declared in
this scope
               LOG (error) << "ky: user bindings: invalid key name: "
<< spec;
^~~~
scons: *** [src/modes/keybindings.o] Error 1
scons: building terminated because of errors.

gauteh:
Nevermind, fixed in: 20b0f68

PR #217, opened 2016-09-12
add install-dir different from prefix

emdete:
to be able to package astroid for voidlinux a distuingished prefix and install/destdir is needed. this patch adds this feature for scons on minimal basis. previous features (prefix==idir) should not be affected.

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -26,6 +26,9 @@ disable_plugins = GetOption ("disable_plugins")
 
 prefix = GetOption ("prefix")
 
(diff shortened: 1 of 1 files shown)


gauteh:
Hi,

I wonder if the install-sandbox option [0] does almost the same? Does it
achieve the necessary stuff for Void Linux?

[0] https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=astroid#n39 (line 39)

Regards, Gaute

M. Dietrich writes on september 12, 2016 11:28:

emdete:
yes it does. thank you!

PR #220, opened 2016-09-15
added ENV=os.environ, which allows it to compile on NixOS

olynch:
I made a change to the SConstruct file which allows it to find the libraries it needs by looking in the user-supplied environment variables instead of hardcoded ones.


gauteh:
olynch writes on september 15, 2016 20:58:
> I made a change to the SConstruct file which allows it to find the libraries it needs by looking in the user-supplied environment variables instead of hardcoded ones.

Looks good, but this makes the environment stuff at lines 127 obsolete,
could you remove those and test?

-gaute

gauteh:
Merged in #263.

PR #221, opened 2016-09-21
check error

emdete:
add error return check and react on it. seen due to the warning from the compiler that the error wasn't checked.


gauteh:
Thanks, merged.

M. Dietrich writes on september 21, 2016 10:13:

Issue #238, opened 2016-12-06
By default: encrypt emaill in answer to encrypted email


hugoroy:
I think this is quite important. If the user receives encrypted email (and that you're able to decrypt it), then his/her reply by default should be encrypted.

gauteh:
Hugo Roy writes on desember 6, 2016 17:33:
> I think this is quite important. If the user receives encrypted email (and that you're able to decrypt it), then his/her reply by default should be encrypted.
> 

Agreed.

gauteh:
Please try #240.

hugoroy:
Maybe I'm wrong but I think it's a good idea to always sign when you encrypt.
Otherwise it's working correctly as far as I can test :-)

gauteh:
Yeah, also notmuch doesn't make the distinction it seems - that might be a bug though.

gauteh:
sign when encrypt. merged and closed.

PR #240, opened 2016-12-07
re: match signature / encryption level to sent e-mail

gauteh:
#238


hugoroy:
Maybe I'm wrong but I think it's a good idea to always sign when you encrypt.
Otherwise it's working correctly as far as I can test :-)

gauteh:
sign when encrypted as well. merged.

PR #243, opened 2016-12-13
In some cases notmuch_thread_get_toplevel_messages and notmuch_messa…

reflectivedevelopment:
…ge_get_replies don't return all the messages in a thread. Lets still load those messages.

I've encountered a weird scenario where I couldn't load some messages in a thread. It appears that the notmuch_thread_get_toplevel_messages and notmuch_message_get_replies don't always return all the messages in a thread.

### DIFF
--- a/src/message_thread.cc
+++ b/src/message_thread.cc
@@ -803,6 +803,39 @@ namespace Astroid {
           add_replies (message, level + 1);
 
         }
+
+        if (messages.size() != (unsigned int) notmuch_thread_get_total_messages (nm_thread))
+        {
+          ustring mid;
+          LOG (error) << "message: thread count not met! Brute force!";
+          for (qmessages = notmuch_thread_get_messages (nm_thread);
+               notmuch_messages_valid (qmessages);
+               notmuch_messages_move_to_next (qmessages)) {
+            bool found;
+            found = false;
+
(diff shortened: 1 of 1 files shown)


gauteh:
Excellent - thanks. Have you notified or discussed this issue with the
notmuch devs? Is this intended or a notmuch bug?

Regards, Gaute

reflectivedevelopment:
I brought it up on slack, but I think I need to write up a bug report. It looks like others are expecting this case to never occur. I'll see if I can get that written up this week.

gauteh:
Kim Ebert writes on desember 15, 2016 18:02:
> I brought it up on slack, but I think I need to write up a bug report. It looks like others are expecting this case to never occur. I'll see if I can get that written up this week.

I plan to include this in the next release, but we should submit a bug
report to notmuch@. If it eventually gets fixed in notmuch we will just
remove it again.

gauteh:
Merged, thanks. Bug report should be submitted to notmuch.

gauteh:
@reflectivedevelopment can you test with your messages about whether this bug still exists? It might have been fixed in notmuch now: https://notmuchmail.org/pipermail/notmuch/2018/026773.html

reflectivedevelopment:
Yeah, I'll give it a shot.

PR #247, opened 2016-12-19
Fix quotes

gauteh:
Replaces #167. Please test.


mxmehl:
I confirm that with this patch I don't see ugly quotes anymore. Thanks!

Issue #244, opened 2016-12-13
theme: could not process scss: Internal Error: File context has no input path


emdete:
since the last debian update astroid can't open the thread view from a index. it terminates with:

[20:03:16.015917] [0x00007f6d78007ac0] [debug] mw: focus-in: 1 active: true, in_quit: false
[20:03:16.873890] [0x00007f6d78007ac0] [debug] ti: open thread: 000000000002511d (true)
[20:03:16.875252] [0x00007f6d78007ac0] [debug] theme: loading..
[20:03:16.875369] [0x00007f6d78007ac0] [info] re: using local resource: /opt/astroid/examples/../ui/thread-view.html
[20:03:16.875494] [0x00007f6d78007ac0] [debug] tv: testing version: 4
[20:03:16.875611] [0x00007f6d78007ac0] [info] re: using local resource: /opt/astroid/examples/../ui/thread-view.scss
[20:03:16.875718] [0x00007f6d78007ac0] [debug] tv: testing version: 4
[20:03:16.876146] [0x00007f6d78007ac0] [info] theme: processing: /opt/astroid/examples/../ui/thread-view.scss
[20:03:16.876469] [0x00007f6d78007ac0] [error] theme: error processing: Internal Error: File context has no input path

(astroid:974): glibmm-ERROR **: 
unhandled exception (type std::exception) in signal handler:
what: theme: could not process scss: Internal Error: File context has no input path

emdete:
it's from libsass. libsass0 3.4.0-1 has the problem, libsass0 3.3.6-2 doesnt. at least i can email again ;)

gauteh:
M. Dietrich writes on desember 13, 2016 20:24:
> it's from libsass. libsass0 3.4.0-1 has the problem, libsass0 3.3.6-2 doesnt. at least i can email again ;)

I can't see that the API has changed. Did you try re-compiling astroid
after the upgrade?

emdete:
sure, even a complete build (after clean) and a git pull. 

no, the api did not change.it seems the behaviour of the lib is to blame. i found the exception in code but did not dive deeper (for example to see if that check was implemented or changed recently)

gauteh:
M. Dietrich writes on desember 14, 2016 8:44:
> sure, even a complete build (after clean) and a git pull. 
> 
> no, the api did not change.it seems the behaviour of the lib is to blame. i found the exception in code but did not dive deeper (for example to see if that check was implemented or changed recently)

[...]

PR #249, opened 2016-12-20
sass: set options directly on file_ctx to avoid options reset

gauteh:
recent libsass (3.4) resets the original context when setting an option
(https://github.com/sass/libsass/issues/2257) causing the input_path to
be lost (#244). here we set the options directly on file_ctx which is a
subclass of options - this might break in the future if file_ctx no
longer sub-classes options. oh well.


emdete:
works for me :)

Issue #245, opened 2016-12-14
Catching Dumb Mistakes Before Sending


Radvendii:
I sent out an email at about 7AM this morning, and I came back at 2PM and it hadn't sent. The reason being that I had forgotten to include a "to" address! Then when I re-sent the message, I forgot the attachments!

So I thought of a feature we could add to astroid that would solve both of these issues. If we could run a script over the message before it sends to check it for possible errors, and then have some confirmation that those are intentional things before sending. So for example checking that there is something in the "to" field, and checking that if there is the word "attachment" or "attached" somewhere in your email, you actually have something attached.

I suggest that there should be some default scripts already turned on (such as checking that there is something in the "to" field but with the option to turn any of these off, and way to add user-defined scripts.

I don't have time right now to look at implementing this, but if people think this is a good idea and nobody wants to do it right now, I can try to implement it later.

gauteh:
Hi, Yes. I think this would be a nice feature. The best might be to
set up a plugin for this so that you could use Python's gmime bindings
to read and check the e-mail.

The method then returns a passed/failed status and an error message with
can be used to describe what the problem is, then we could present that
to ask whether the user wants to send the e-mail still. What do you
think?

Defaults, like you suggested, could be tested and implemented anyway.

If you have the chance to work on a patch for this that would be great.

mxmehl:
+1. With Thunderbird I liked the attachment reminder. You're able to define specific words (e.g. attachment, attached, anhang, anbei) and if one of those words is in the mail, the clients asks you whether you really want to send the mail without any attachment (if there isn't any).

I agree that having this as a plugin would be the best solution to keep astroid slim. I would be happy to test your implementations!

kirschner:
* mxmehl <notifications@github.com> [2016-12-23 06:55:59 -0800]:

> +1. With Thunderbird I liked the attachment reminder. You're able to
> define specific words (e.g. attachment, attached, anhang, anbei) and
> if one of those words is in the mail, the clients asks you whether you
> really want to send the mail without any attachment (if there isn't
> any).

That would be really good. 

Maybe it helsp; in mutt you can use a vim plugin for that:
http://www.vim.org/scripts/script.php?script_id=1015

gauteh:
Matthias Kirschner writes on desember 23, 2016 16:48:
> * mxmehl <notifications@github.com> [2016-12-23 06:55:59 -0800]:
> 
>> +1. With Thunderbird I liked the attachment reminder. You're able to
[...]

PR #252, opened 2016-12-24
em: catch missing fields and mentions of attachments before sending

gauteh:
#245

### DIFF
--- a/src/config.cc
+++ b/src/config.cc
@@ -185,6 +185,8 @@ namespace Astroid {
     default_config.put ("editor.charset", "utf-8");
     default_config.put ("editor.save_draft_on_force_quit", true);
 
+    default_config.put ("editor.attachment_words", "attach");
+
     /* mail composition */
     default_config.put ("mail.reply.quote_line", "Excerpts from %1's message of %2:"); // %1 = author, %2 = pretty_verbose_date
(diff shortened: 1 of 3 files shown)


mxmehl:
Thanks! But how can I define multiple words? My config looks like this:

`"attachment_words": "attachment,attached,anbei,anhang,angehängt"`

I only get the error message when using the whole string in the mail. Content like `see my attachment here` doesn't trigger a warning with this config. I also tried semicolons.

The warning when using no `To` worked fine for me.

gauteh:
Thanks for testing. 

I changed it to use ',' as separator, please try again. Also, incomplete
words will be matched. So just using 'attach' rather than attachment and
attached will be faster.

mxmehl:
Much better, thanks!

My last wish: could you please make it case-insensitive?

gauteh:
Max writes on desember 24, 2016 16:55:
> Much better, thanks!
> 
> My last wish: could you please make it case-insensitive?

Should work now!

mxmehl:
>> My last wish: could you please make it case-insensitive?
>
> Should work now!

It does! Thanks for the christmas present ;)

gauteh:
Max writes on desember 24, 2016 22:50:
>>> My last wish: could you please make it case-insensitive?
>>
>> Should work now!
> 
> It does! Thanks for the christmas present ;)
> 

Merged. merry christmas :)

PR #258, opened 2017-01-03
config: Respect $NOTMUCH_CONFIG for notmuch's config file

lucc:
The environment variable is documented in the main notmuch man page.

### DIFF
--- a/src/config.cc
+++ b/src/config.cc
@@ -111,7 +111,13 @@ namespace Astroid {
   ptree Config::setup_default_config (bool initial) {
     ptree default_config;
     default_config.put ("astroid.config.version", CONFIG_VERSION);
-    std::string nm_cfg = path(std_paths.home / path (".notmuch-config")).string();
+    const char * nm_cfg_env = getenv ("NOTMUCH_CONFIG");
+    string nm_cfg;
+    if (nm_cfg_env == NULL) {
+      nm_cfg = path (std_paths.home / path (".notmuch-config")).string ();
+    } else {
+      nm_cfg = path (nm_cfg_env).string ();
+    }
     default_config.put ("astroid.notmuch_config" , nm_cfg);
 
     default_config.put ("astroid.debug.dryrun_sending", false);


gauteh:
Thanks, this will only use $NOTMUCH_CONFIG for the default setup though.

If we want the environment variable to have effect also after a
astroid-config exists we should change how the notmuch-config value is
parsed in the first place: perhaps have a config value of 'default' mean
to either use the env variable or use ~/.notmuch-config:

  astroid.notmuch_config = 'default'

I think for now we should go for this option, what do you think?

One problem that the current configuration system suffers from is that
it does not distinguish between default and unset values. So that
config values from earlier versions are tedious to update.

Regards, Gaute

lucc:
Sorry I did not have a look at the whole config handling system in astroid. I just installed astroid after you mentioned the release on the notmuch mailinglist yesterday. And then I got the error message because I use $NOTMUCH_CONFIG and thought I'd send you a fix.

What would happen if a user had a config file named "default" and wanted to use that in the case you suggested?

Here is what I know of other applications and how they handle the potential different sources of config options or at least what I expect how settings should normally be loaded from different sources (python like pseude code):
```python
args = parse_command_line()
if args.config_file:
  config_file = args.config_file
else if environment.config_file:
  config_file = environment.config_file
else
  config_file = application_defaults.config_file
config_file.load()
# for some option "foo" one would now do
if args.foo:
  foo = args.foo
else if environment.foo:
  foo = environment.foo
else if config_file.foo:
  foo = config_file.foo
else:
  foo = application_defaults.foo
```
(Obviously not all options will have corresponding command line switches or environment variables but you get the idea I hope)

The `config_file` above would be the main astroid config file (default at `~/.config/astroid/config`). I am not sure if the notmuch config file should be handled like the default config file or rather like all other options. I think rather like all other options as it seems possible to set its path in the main config file. Maybe the logic for all further options would have to be extended to check the notmuch config file after the astroid config file and before using the application defaults (if options can also be set in the notmuch config file).

gauteh:
Yes - there are a few options. I think for now we should just go for the
'default' way: 

It is unlikely that the user will have a file called default, anyway,
the user can always specify the full path and it should be fine.

Currently, all configuration options that are common between notmuch and
astroid are gathered from the notmuch config file only.

I think if you move your logic to line 303 (read notmuch config) and
adapt it a little (do a check on the astroid.notmuch_config option) and
change the default config option of that value to 'default' we're good
for now. A complete change of the config system is a bigger job.

gauteh:
Implemented in #543.

PR #263, opened 2017-01-11
Update SConstruct to allow compiling on NixOS

SuprDewd:
This subsumes #220, additionally removing some redundant work with `os.environ`. Has (only) been tested on NixOS.

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -5,7 +5,7 @@ from subprocess import *
 def getGitDesc():
   return Popen('git describe --abbrev=8 --tags --always', stdout=PIPE, shell=True).stdout.read ().strip ()
 
-env = Environment ()
+env = Environment (ENV=os.environ)
 
 AddOption ("--release", action="store", dest="release", default="git", help="Make a release (default: git describe output)")
 AddOption ("--enable-debug", action="store", dest="debug", default=None, help="Enable the -g flag for debugging (default: true when release is git)")
--- a/SConstruct
+++ b/SConstruct
@@ -5,8 +5,6 @@ from subprocess import *
 def getGitDesc():
   return Popen('git describe --abbrev=8 --tags --always', stdout=PIPE, shell=True).stdout.read ().strip ()
 
-env = Environment (ENV=os.environ)
-
 AddOption ("--release", action="store", dest="release", default="git", help="Make a release (default: git describe output)")
 AddOption ("--enable-debug", action="store", dest="debug", default=None, help="Enable the -g flag for debugging (default: true when release is git)")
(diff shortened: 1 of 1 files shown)


gauteh:
Thanks, will test here on Arch, and comment back.

gauteh:
One concern about this and #220, based on: http://scons.org/doc/HTML/scons-man.html

If we copy the whole environment we will get builds that are quite dependent on the user configuration. Is there a lot of extra variables that are required by NixOS or just a few?

SuprDewd:
It seems to be `PATH`, `PKG_CONFIG_PATH`, `XDG_DATA_DIRS`, `NIX_LDFLAGS` and `NIX_CFLAGS_COMPILE`, the last two being NixOS-specific.

After digging into some of these, at least one big issue is with `NIX_LDFLAGS`. It's similar to `LDFLAGS`, but also contains some stuff that isn't meant for `LDFLAGS`, like `-rpath`, which should be passed directly to `ld`, or as `-Wl,-rpath` to `gcc`. NixOS has a wrapper for `ld` that takes care of all of this.

This is something that can be fixed in `SConstruct`, but is starting to become way too NixOS-specific. So here are two options:

1) We abandon this and #220 as a wontfix and make a note of it in the Wiki. As for the astroid package in the NixOS package repository, it's already applying this as a [patch](https://github.com/NixOS/nixpkgs/blob/master/pkgs/applications/networking/mailreaders/astroid/propagate-environment.patch), and as far as I can see, this is how other NixOS packages have been dealing with scons. That make sense, as builds in NixOS already have a sanitized environment.
2) If you don't mind soiling `SConstruct` with platform-specifics, then we could add the five variables mentioned at the top to the environment when they are present.

Which do you prefer?

gauteh:
Bjarki Ágúst Guðmundsson writes on januar 12, 2017 22:34:
> It seems to be `PATH`, `PKG_CONFIG_PATH`, `XDG_DATA_DIRS`, `NIX_LDFLAGS` and `NIX_CFLAGS_COMPILE`, the last two being NixOS-specific.
>
> [...]
>
> Which do you prefer?

Hm, I'd be happy with going for a '--propagate-environment' option
which then implement what this patch does. Would that work for you? It
seems like a better option than 2) with adding some very specific
variables, and might work for other scenarios as well.

In this case I think that the variables that already are captured can
stay since they are fairly standard. What do you think?

SuprDewd:
Sounds good. Updated the commit.

gauteh:
Bjarki Ágúst Guðmundsson writes on januar 14, 2017 0:32:
> Sounds good. Updated the commit.

Thanks, merged.

Issue #260, opened 2017-01-04
Styling flagged items in thread index.


hughwilliams94:
Might there be some way to change the appearance of flagged messages to highlight them a little, perhaps adding a slight background, or a different font colour a la sup? I have looked in the config but the options there don't fit this purpose

gauteh:
Hi, this is currently not possible. It could be done with a plugin though in a similar way as it is done in sup. It is already possible to style the tags (though not based on flagged status), and this could work similarly.

We'd end up with lots of configuration options for the different cases, but that might be the easiest way for people to style the list entry.

gauteh:
Related: #219.

PR #268, opened 2017-02-05
testing more compact thread view see issue #180

hugoroy:
See issue #180


gauteh:
Nice, thanks - perhaps an excerpt of the subject can be moved up after the author name?

gauteh:
And maybe tags below the date ( or is that used for the icons already? )

hugoroy:
↪ Gaute Hope, février 17, 2017 8:09 :
> Nice, thanks - perhaps an excerpt of the subject can be moved up after the author name?

The reason I did not include the Subject is because, usually, there is
one subject per thread. So you already see the Subject in the index-view
as well as in the tabs. Repeating the Subject over and over again in
each message preview seemed too much. The subject still shows himself
for unread/expanded emails, so I think it's safe to avoid it here.

hugoroy:
↪ Gaute Hope, février 17, 2017 8:10 :
> And maybe tags below the date ( or is that used for the icons already? )

Yes, that's where the attachment icon is.

I tried various things, not always succesful :^)

hugoroy:
↪ Gaute Hope / février 20, 2017 10:50:
> We could maybe do some comparison on the subjects and check to what
> degree it matches the previous (discarding any 'Re:' and 'Fwd:'s) and if
> it is below a certain treshold it could be included. This could be
> implemented by adding a CSS tag to that message.

This seems complicated for not much benefit (again, when the message is
expanded, you can see the subject).

> Another option is to try and detect whether this is a patch (already do
> to some degree) and add a .patch CSS tag to all the messages in that case.

That could be a good default for developers. Easy to implement if
there's already an HTML class attached automatically!

gauteh:
Hugo Roy writes on mars 2, 2017 0:42:
> ↪ Gaute Hope / février 20, 2017 10:50:
>> We could maybe do some comparison on the subjects and check to what
>> degree it matches the previous (discarding any 'Re:' and 'Fwd:'s) and if
>> it is below a certain treshold it could be included. This could be
>> implemented by adding a CSS tag to that message.
> 
> This seems complicated for not much benefit (again, when the message is
> expanded, you can see the subject).

Would it be hard to make your current modifications conditional on a CSS
tag? So that it could be toggled with a keybinding, or that this issue
could be implemented in the future?

>> Another option is to try and detect whether this is a patch (already do
>> to some degree) and add a .patch CSS tag to all the messages in that case.
> 
> That could be a good default for developers. Easy to implement if
> there's already an HTML class attached automatically!

Alright, I'll add the .patch tag to detected messages.

hugoroy:
↪ Gaute Hope / mars 2, 2017 9:19:
> Hugo Roy writes on mars 2, 2017 0:42:
>> ↪ Gaute Hope / février 20, 2017 10:50:
>>> We could maybe do some comparison on the subjects and check to what
>>> degree it matches the previous (discarding any 'Re:' and 'Fwd:'s) and if
>>> it is below a certain treshold it could be included. This could be
>>> implemented by adding a CSS tag to that message.
>> 
>> This seems complicated for not much benefit (again, when the message is
>> expanded, you can see the subject).
> 
> Would it be hard to make your current modifications conditional on a CSS
> tag?

No, it would be easy. The hard part definitely is on your end :-) 

> So that it could be toggled with a keybinding, or that this issue
> could be implemented in the future?

>>> Another option is to try and detect whether this is a patch (already do
>>> to some degree) and add a .patch CSS tag to all the messages in that case.
>> 
>> That could be a good default for developers. Easy to implement if
>> there's already an HTML class attached automatically!
> 
> Alright, I'll add the .patch tag to detected messages.


IIRC, if a given CSS tag is detected (e.g., ".patch" or
".different-subject") meaning that the subject is different from the
previous message, then I would not hide the subject from the email
preview.

gauteh:
Hugo Roy writes on mars 2, 2017 9:32:
> ↪ Gaute Hope / mars 2, 2017 9:19:
>> Hugo Roy writes on mars 2, 2017 0:42:
>>> ↪ Gaute Hope / février 20, 2017 10:50:
>>>> We could maybe do some comparison on the subjects and check to what
>>>> degree it matches the previous (discarding any 'Re:' and 'Fwd:'s) and if
>>>> it is below a certain treshold it could be included. This could be
>>>> implemented by adding a CSS tag to that message.
>>> 
>>> This seems complicated for not much benefit (again, when the message is
>>> expanded, you can see the subject).
>> 
>> Would it be hard to make your current modifications conditional on a CSS
>> tag?
> 
> No, it would be easy. The hard part definitely is on your end :-) 
> 
>> So that it could be toggled with a keybinding, or that this issue
>> could be implemented in the future?
> 
>>>> Another option is to try and detect whether this is a patch (already do
>>>> to some degree) and add a .patch CSS tag to all the messages in that case.
>>> 
>>> That could be a good default for developers. Easy to implement if
>>> there's already an HTML class attached automatically!
>> 
>> Alright, I'll add the .patch tag to detected messages.
> 
> IIRC, if a given CSS tag is detected (e.g., ".patch" or
> ".different-subject") meaning that the subject is different from the
> previous message, then I would not hide the subject from the email
> preview.
> 

'patch' is now added to the email CSS class list for emails that appear
to be patches.

gauteh:
Gaute Hope writes on mars 2, 2017 15:00:
>> IIRC, if a given CSS tag is detected (e.g., ".patch" or
>> ".different-subject") meaning that the subject is different from the
>> previous message, then I would not hide the subject from the email
>> preview.
> 
> 'patch' is now added to the email CSS class list for emails that appear
> to be patches.
> 

'different_subject' is now added for all messages deviating from the
subject of the first thread. That is subject lines that contain anything
else than the original subject and additional 'Re:'s and white-spaces.

hugoroy:
@gauteh does it work?

mxmehl:
Thanks for your discussion so far. However, my needs are different from yours, so I modified my `thread-view.scss` a bit by adding:

```css
body:not(.nohide) .email.hide .header_container .preview {
  display: none !important;
}
body:not(.nohide) .email.hide .header_container {
  padding: 2px !important;
}
body:not(.nohide) .email.hide .header_container .attachment.icon {
  float: left;
  height: 12px;
}
body:not(.nohide) .email.hide .header_container .subject, body:not(.nohide) .email.hide .header_container .tags {
  float: left;
}
```

One - a bit OT - question remains: Can I make the subjects of unread mails bold? Is there a class for it?

gauteh:
Max writes on mars 13, 2017 13:36:
> One - a bit OT - question remains: Can I make the subjects of unread mails bold? Is there a class for it?
> 

Good point, is trivial to add - but need to make sure this is kept in
sync when unread tag is added/removed.

gauteh:
Gaute Hope writes on mars 13, 2017 13:40:
> Max writes on mars 13, 2017 13:36:
>> One - a bit OT - question remains: Can I make the subjects of unread mails bold? Is there a class for it?
>> 
> 
> Good point, is trivial to add - but need to make sure this is kept in
> sync when unread tag is added/removed.

I've now added the 'unread' css tag as well, these are now also updated
on tag changes.

mxmehl:
> I've now added the 'unread' css tag as well, these are now also updated on tag changes.

Nice, works like a charm! I've implemented it in my CSS:

```css
body:not(.nohide) .email.hide.unread {
  background-color: #dedfff !important;
}
body:not(.nohide) .email.hide.unread .header_container .subject {
  font-weight: 700;
}
```

By the way, would a CSS class like unread also be possible for "flagged"? Perhaps related to #260

gauteh:
Max writes on mars 15, 2017 15:52:
>> I've now added the 'unread' css tag as well, these are now also updated on tag changes.
> 
> Nice, works like a charm! I've implemented it in my CSS:
> 
> ```css
> body:not(.nohide) .email.hide.unread {
>   background-color: #dedfff !important;
> }
> body:not(.nohide) .email.hide.unread .header_container .subject {
>   font-weight: 700;
> }
> ```
> 
> By the way, would a CSS class like unread also be possible for "flagged"? Perhaps related to #260 
> 

I was briefly considering just adding all tags as css tags - need some
special astroid tags as well, would this be too messy?

mxmehl:
Am 15.03.2017 um 15:53 schrieb Gaute Hope:
>> By the way, would a CSS class like unread also be possible for
> "flagged"? Perhaps related to #260
>>
> 
> I was briefly considering just adding all tags as css tags - need some
> special astroid tags as well, would this be too messy?

Probably yes... How about making it configurable which tags are being
set in a CSS class? Or would this in turn make the config file much more
confusing (@hugoroy had some concerns about this AFAIR).

hugoroy:
The config seems unnecessary.

However it makes sense to expose the notmuch tags as HTML classes for the CSS.

To avoid any perturbation we should use a prefix for this and be able to handle weird characters , e.g. for the tag "list/astroid"

    class="notmuchtag-list-astroid"


Le 15 mars 2017 15:57:50 GMT+01:00, Max <notifications@github.com> a écrit :
>Am 15.03.2017 um 15:53 schrieb Gaute Hope:
>>> By the way, would a CSS class like unread also be possible for
>> "flagged"? Perhaps related to #260
>>>
>> 
>> I was briefly considering just adding all tags as css tags - need
>some
>> special astroid tags as well, would this be too messy?
>
>Probably yes... How about making it configurable which tags are being
>set in a CSS class? Or would this in turn make the config file much
>more
>confusing (@hugoroy had some concerns about this AFAIR).

-- 
Envoyé de mon appareil Android avec K-9 Mail. Veuillez excuser ma brièveté.

gauteh:
Hugo Roy writes on mars 15, 2017 16:51:
> The config seems unnecessary.
> 
> However it makes sense to expose the notmuch tags as HTML classes for the CSS.
> 
> To avoid any perturbation we should use a prefix for this and be able to handle weird characters , e.g. for the tag "list/astroid"
> 
>     class="notmuchtag-list-astroid"

Yes, it needs to be sanitized. The only argument against a prefix is
that it is less intuitive, but that might not be a real problem.

gauteh:
Gaute Hope writes on mars 15, 2017 18:47:
> Hugo Roy writes on mars 15, 2017 16:51:
>> The config seems unnecessary.
>> 
>> However it makes sense to expose the notmuch tags as HTML classes for the CSS.
>> 
>> To avoid any perturbation we should use a prefix for this and be able to handle weird characters , e.g. for the tag "list/astroid"
>> 
>>     class="notmuchtag-list-astroid"
> 
> Yes, it needs to be sanitized. The only argument against a prefix is
> that it is less intuitive, but that might not be a real problem.
> 

All notmuch tags are now added with the 'nm-' prefix, so 'unread' is
now: 'nm-unread'. They are escaped using Glib::Markup::escape_text and /
is replaced with -.

mxmehl:
How should we proceed here? 

It seems our needs for a "compact thread view" are diverging. Shall we just document our additional CSS parts in the wiki, e.g. here: https://github.com/astroidmail/astroid/wiki/Customizing-the-user-interface ?

gauteh:
Max writes on mars 16, 2017 11:51:
> How should we proceed here? 
> 
> It seems our needs for a "compact thread view" are diverging. Shall we just document our additional CSS parts in the wiki, e.g. here: https://github.com/astroidmail/astroid/wiki/Customizing-the-user-interface ?
> 

I think putting more emphasis on unread messages could also be included
in astroid. What do you think Hugo?

mxmehl:
> I think putting more emphasis on unread messages could also be included in astroid.

Do you mean only in TV like marking them as I did or in general? Because in general I'd also find one or two occasions where unread messages could be highlighted. Should I open a new ticket for that?

gauteh:
Max writes on mars 16, 2017 12:37:
>> I think putting more emphasis on unread messages could also be included in astroid.
> 
> Do you mean only in TV like marking them as I did or in general? Because in general I'd also find one or two occasions where unread messages could be highlighted. Should I open a new ticket for that?

What do you mean in general? In thread-index?

mxmehl:
> What do you mean in general? In thread-index?

Just for example I could think of marking tabs with unread mails bold, or in Saved Searches make unread searches/queries bold. TI seems to be clear enough.

gauteh:
Max writes on mars 16, 2017 18:41:
>> What do you mean in general? In thread-index?
> 
> Just for example I could think of marking tabs with unread mails bold, or in Saved Searches make unread searches/queries bold. TI seems to be clear enough.

Ah, I see - saved searches I agree. I think bold tab headers might steal
too much attention? Would be nice with a PR to test :)

gauteh:
I think this one will have to wait for the next release.

hugoroy:
Let's focus this one the thread view and open another one for  the thread-index and other modifications. They seem quite independent one from another to me.

I agree that now that we have nm- tags as css classes, we can do some very nice stuff for standard tags like unread, but also flagged, encrypted, signed, etc.

I will make a separate pull request to come up with a coherent proposal, and I will test the compact thread view here (i.e. when to hide the "Subject" of a non-expanded email, etc.).

Does that sound like a good plan?

gauteh:
Hugo Roy writes on mars 17, 2017 17:15:
> Let's focus this one the thread view and open another one for  the thread-index and other modifications. They seem quite independent one from another to me.
>
> [..]
>
> Does that sound like a good plan?

Excellent!

gauteh:
Closing, since this would need to be re-implemented.

PR #274, opened 2017-02-08
Fix spelling of astroid

SuprDewd:
s/astorid/astroid/g


gauteh:
Thans, merged. Updated compose to composes as well.

Issue #251, opened 2016-12-23
Show which pane is focused


mxmehl:
I started to use astroid's panes more often, especially opening threads in panes. However, I have noticed two unwanted behaviours:

1. Switching between panes is possible with C-Tab. But if I want to switch from right (thread view) to left (thread index), I have to press C-Tab twice. The other way round, pressing it once is sufficient.

2. There is no way how to tell which pane is active currently. When opening a thread in a new pane, it is the right pane by default, but as soon as I start to work in the other pane or different tabs, I start to get confused. Can we mark the active pane optically, for example by a horizontal blue bar on the bottom or top of the active pane, equivalently to how we mark active messages in thread view?

gauteh:
Hi,

there was a discussion earlier were I was considering dropping the
panes. I no longer use them myself, and tracking focus between them has
proven a bit tricky.

The reasoning for dropping them is that the functionality can
essentially be achieved by using two windows (you can open a message
directly in a new window using: C-Return).

more below...:


mxmehl writes on desember 23, 2016 16:47:
> I started to use astroid's panes more often, especially opening threads in panes. However, I have noticed two unwanted behaviours:
> 
> 1. Switching between panes is possible with C-Tab. But if I want to switch from right (thread view) to left (thread index), I have to press C-Tab twice. The other way round, pressing it once is sufficient.

Usually it should be M-Tab, C-Tab is probably a Gtk binding for focusing
the previous element. That is why it behaves a bit unlogical. Does M-Tab
work ok?

 
> 2. There is no way how to tell which pane is active currently. When opening a thread in a new pane, it is the right pane by default, but as soon as I start to work in the other pane or different tabs, I start to get confused. Can we mark the active pane optically, for example by a horizontal blue bar on the bottom or top of the active pane, equivalently to how we mark active messages in thread view?

Yeah - that's a good point.

mxmehl:
Am 24.12.2016 um 16:52 schrieb Gaute Hope:
> there was a discussion earlier were I was considering dropping the
> panes. I no longer use them myself, and tracking focus between them has
> proven a bit tricky.
> 
> The reasoning for dropping them is that the functionality can
> essentially be achieved by using two windows (you can open a message
> directly in a new window using: C-Return).

Hm, I wanted to start using panes on my large 16:9 screen because I'm
used to having all content in one window (except the new message
editor). But if this causes too much maintenance work, I could live with
dropping panes.

[...]

Issue #269, opened 2017-02-06
Display search results as single messages


mxmehl:
It would be awesome to optionally see the results of search queries as individual messages instead of the threads containing a matching message. This would come in handy when you have multiple matching giant mail threads and look for a single message in them with multiple queries, for example `tag:flagged AND from:someguy AND contact`.

notmuch allows to see single messages' ids by adding `--output=messages` to the search query. But astroid doesn't seem to understand such options.

gauteh:
Related to #100.

gauteh:
I would like this too, it would be good to be able to toggle the
thread-index between thread and message view (either with a tree view
of the messages, or flat).

Max writes on februar 6, 2017 19:25:

mxmehl:
> I would like this too, it would be good to be able to toggle the thread-index between thread and message view (either with a tree view of the messages, or flat).

That would be nice as well!

hugoroy:
That's definitely an interesting alternative use. I hope it won't be too much effort to implement or require strange tweaks in the rest of astroid!

kirschner:
I was also just trying to achieve that and now found this issue. If it would be possible to toggle between thread and message view, that would be very helpful. Thanks for considering it.

Issue #250, opened 2016-12-22
Improve patch applying and maintaining


gauteh:
I want to improve the way patches can be applied. Currently you can yank the content (raw or decoded), or the message-id. I think with some improvements various workflows could be supported. The improvements I am thinking about are:

* Make a run hook in thread_view that can take one or more message-id's, this can be used with a script that gets the filenames using notmuch and runs `git am` on them. You could have the script do different things depending on what tags the message has, and even add a tag to the message afterwards for your own records.
* Make a tree-view in the thread_index (#100, #180) which supports the same behavior for running hooks

Current, relevant, possibilities include:
* Copying message content
* Copying message-id
* Saving one or more messages

Eventually, there should be some suggested workflows for how to do this efficiently in the wiki. Does anyone have any experience or particular workflow they would like to see possible?

gauteh:
With the latest improvements to Yank raw and to #181 it is now smoother to apply patchsets. 

By 'marking' several messages in thread-view and using the multi-key + `Y` (yank raw) you get a mailbox synposis of the marked messages copied to the primary clipboard: https://github.com/astroidmail/astroid/commit/5fd2c763fbabdc9f9c6ea56d6c7d88d0425db128 explains a possible workflow:


Yank raw on multiple-patches produces a mailbox-like (without >From
filtering) synthesis of the patches, suitable for `git am`. So to apply
a patch set sent with git-send-email or similar do:

* open the thread
* mark the messages that are part of the patch (or use 'T' to mark all)
* open the multi-key
* use 'Y' to yank raw, the mbox is now in the primary clipboard
* open a terminal and navigate to your repository
* execute `git am` or `git am -3`
* press Shift-Insert (or your terminals equivalent) to paste from
  primary clipboard.
* press Ctrl+D to stop pasting

You can also apply individual patches by using 'Y' on a message outside
multi-key mode (not part).

Inline patches can mostly be copied with 'y' (which yanks decoded text)
and then be applied using `git apply`.

> This will go into a wiki page on using astroid for appyling patches.

gauteh:
https://github.com/astroidmail/astroid/wiki/Working-with-patches-in-Astroid

gauteh:
@gregkh: You might be interested.

gregkh:
@gauteh oh nice!  I'll have to try this out soon, thanks for the ping.

gauteh:
Greg Kroah-Hartman writes on februar 25, 2017 10:50:
[...]

PR #280, opened 2017-02-16
Pane show focus

gauteh:
#251 and move pane swap to M-space

### DIFF
--- a/History.txt
+++ b/History.txt
@@ -9,6 +9,9 @@
 
 * `scons --propagate-environment` uses environment from build environment.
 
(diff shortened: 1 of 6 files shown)


mxmehl:
Works for me. I'll do a bit more testing during the next days. I'm not sure yet whether I'll use the panes extensively but these changes make it much more comfortable.

But we have to mention that the pane switch keybinding and its descriptor changed (at least for me).

gauteh:
Right, I forgot I changed that. Will go in History.txt.

gauteh:
I plan on using the panes more for #250 and #269.

Issue #261, opened 2017-01-06
Reply-To causes buggy replies-to-all


mxmehl:
It seems replying to mails containing the Reply-To header causes some strange behaviour of astroid. I couldn't exactly reproduce it with my own tests but they caused another (edge-case) flaw:

Header 1 (shortened):
```
To: xxyyzz@q.fsfe.org
From: ****** <****@fsfe.org>
Reply-To: xxyyzz@q.fsfe.org
In-Reply-To: <1483715512.******.max.****@fsfe.org>
Precedence: bulk
X-ML-Address: xxyyzz@q.fsfe.org
```

Pressing Reply-all (my default on mailing lists) causes astroid to use `xxyyzz@q.fsfe.org` *twice* as recipient of the new mail. Simple Reply work correctly, using the xxyyzz address.

Header 2 (shortened):
```
From: Max Mehl <****@mehl.mx>
Reply-To: "max.****@fsfe.org" <max.****@fsfe.org>
To: "max.****@fsfe.org" <max.****@fsfe.org>
```

Reply-all causes astroid to use no recipient at all. Simple reply again works as expected.

gauteh:
Uh, oh, yeah. That doesn't look right...

gauteh:
Please try #281

PR #281, opened 2017-02-16
re, address: remove duplicate addresse when reply-alling

gauteh:
#261


mxmehl:
Thanks, works for me.

gauteh:
Merged.

Issue #241, opened 2016-12-07
gpg: too many open files / high cpu usage


emdete:
i encounter the error "too many open files" quite often since i use gpg with every email.

maybe something's broken in the usage of gpg?

gauteh:
Have you confirmed this to be only with GPG? Does this happen when you have astroid running for a very long time? Do you have any logs?

Does astroid leave any gpg instances open?

gauteh:
M. Dietrich writes on desember 7, 2016 17:20:
> i encounter the error "too many open files" quite often since i use gpg with every email.

Test latest master, I had forgotten to unref the gpg context.

gauteh:
Closing, please re-open if issue persists.

hugoroy:
Astroid seems to leave a lot of gpg instances open, yes (and gpg hangs at 100%CPU since yesterday)

gauteh:
Hugo Roy writes on desember 9, 2016 16:10:
> Astroid seems to leave a lot of gpg instances open, yes (and gpg hangs at 100%CPU since yesterday)

Still the case after 4e523d6f375d4050e5a53143885f7712f5bacf91?

hugoroy:
↪ Gaute Hope, décembre 9, 2016 4:21 :
> Hugo Roy writes on desember 9, 2016 16:10:
>> Astroid seems to leave a lot of gpg instances open, yes (and gpg hangs at 100%CPU since yesterday)
> 
> Still the case after 4e523d6f375d4050e5a53143885f7712f5bacf91?
> 

Yes, just recompiled master with the AUR.

Actually now, as soon as astroid starts, it spawns a gpg2 and seahorse
(which apparently are hanging on something at 100%CPU).

gauteh:
I don't see how that can happen without opening an e-mail thread? Is it something that goes though the polling script and notmuch indexing of encrypted e-mails?

hugoroy:
I downgraded gnupg. Not sure what the problem is exactly, but it doesn't seem to be astroid :)

gauteh:
What is the problematic gnupg version? We should check if other notmuch users have the same problem, otherwise maybe report a gmime / gnupg bug.

hugoroy:
2.1.16. 

I unfortunately won't have the time to dig further :)

gauteh:
@emdete can you confirm whether your bug is either fixed by 4e523d6 or related to the gnupg version?

gauteh:
[...]

PR #286, opened 2017-02-21
Crypto too many files

gauteh:
#241


gauteh:
If you check out 22e0543 you should be able to confirm that the tests fail.

Issue #292, opened 2017-02-25
Editing the "from" field in the editor and lose the content


hugoroy:
If I write an email and try to change the 'from' field in the editor itself, I lose the content of the email I juste wrote when I quit the editor.

gauteh:
Hugo Roy writes on februar 25, 2017 20:08:
> If I write an email and try to change the 'from' field in the editor
> itself, I lose the content of the email I juste wrote.

Oh, that's bad. Is this with the external editor? Is the new from field
value a valid e-mail? (one that is registered as an account in the
astroid config).

hugoroy:
↪ Gaute Hope / février 25, 2017 22:29:
> Oh, that's bad. Is this with the external editor?

Yes, emacs in my case.

> Is the new from field value a valid e-mail? (one that is registered as
> an account in the astroid config).

Yes it is valid. Yes, it is also an account.


The problem is, I can't always produce this bug, but I've noticed that
that sometimes happen (which is why I have grown a habit of being
careful with editing the "From" field from the editor... but I sometimes
still do it).

gauteh:
Does this only happen when any of the involved accounts has a signature?

hugoroy:
It does not seem to be related to that.

But it seems I can always produce the bug when I am writing a New email (as opposed to replying to an existing email).

gauteh:
I cannot reproduce this. Could you post a log with a minimum of other stuff going on?

hugoroy:
↪ Gaute Hope / février 26, 2017 17:24:
> I cannot reproduce this. Could you post a log with a minimum of other stuff going on?


So here's what I did.

 1. Write a new mail (not replying to an existing email).
 2. Edit the From field (showing the default email address) by replacing
 with another (valid) email address used for another account - Edit the
 email subject and the email body, save and quit Emacs editor
 3. The From address has changed appropriately but the subject and
 content of the email is lost.
 4. Then I quit.


Here's the full log (If that helps, I uploaded my astroid config to <https://github.com/hugoroy/.astroid>).




    [18:59:41.558526] [0x00007ff634713340] [debug] mw: focus-in: 1 active: true, in_quit: false
    [18:59:42.123624] [0x00007ff634713340] [info] re: using installed resource: /usr/share/astroid/ui/edit-message.glade
    [18:59:42.126287] [0x00007ff634713340] [info] em: msg id: 1488131982.lgiyq7ixeu.hugo@local.fqdn
[...]

PR #293, opened 2017-02-28
#292: em: do not write a new draft when in reading

gauteh:
#292


gauteh:
@hugoroy: This one needs some stress-testing with replying, forwarding, changing from addresses, reply to addresses, external internal editor etc.

Issue #294, opened 2017-03-01
Add Reply-to-list option


mxmehl:
Many MUAs have the ability to make a reply-to-list. Currently, when replying to a message on a mailing list (I've tested Mailman), astroid replies to the list **and** the sender of the mail I reply to.

Many people don't like this behaviour. So I'd like to ask for adding a reply-to-list alongside the already existing simple reply and reply-to-all, including a configurable shortcut of course. 

I guess, astroid should respect the `List-*` headers therefore, like `List-Post: <mailto:abc@lists.fsfe.org>`. 

Details:
- If such headers do not exist, and the user uses reply-to-list anyway, reply-to-all should be chosen instead.
- Other `To` and `Cc` values should still be respected. Only the sender should be left out (I guess this is the logic other MUAs use).

kirschner:
I'd like to second this. List-reply is really one of the most crucial features for me as it is considered impolite on many mailing lists if you additionally include the sender. 

Thank you for considering it!

hugoroy:
I think there are two things here:

1. Ensure that the "Astroid Reply-To Suggestions" work correctly. See the bottom right-hand corner screenshot which has a drop-down list of available Astroid Reply-To Suggestions: https://raw.githubusercontent.com/astroidmail/astroid/master/doc/full-demo-external.png 


    However in my experience, I often have to manually make custom edits to the Reply-To. I find that sometimes the Astroid Reply-To Suggestions do not work as expected. Am I the only one? Should I file a separate issue for these?


2. Provide some keybindings when replying on an email which will select the desired Astroid Reply-To Suggestion. For instance, here are some suggestions of keybindings from the top of my head which would need to be checked:

    - "r" should follow the "Reply-to:" field given by the email responding to; 
    - "g" should override the "Reply-to:" field and select the group of all participants in the email instead; 
    - "L" should override the "Reply-to:" field and select the "List-*" field, 
    - "R" should override the "Reply-to:" field and select the sender only instead;
    - "c" should be like "R" but in a new thread.

I think these above already exist but maybe ensuring that this is consistent and coherent could be good. I will try to make a list of what the current default keybindings for these are, if they exist.

gauteh:
`r`, `g` already works as you describe. `c` composes a new e-mail to the sender, like you describe but not to reply-to.

gauteh:
Check out #297. There's also a config option for not including the sender in to (by default included).

gauteh:
Closed in #297

[...]

PR #297, opened 2017-03-02
Re mailinglist

gauteh:
#294 

Use `M` to reply to mailinglist and `R` to reply to sender. Set the config option `mail.reply.mailinglist_reply_to_sender` to `false` to not include sender in To field when replying to mailinglist.


mxmehl:
Great, thanks! I ran some tests and it worked out just fine until now.

PR #301, opened 2017-03-02
hint: how to cancel when the message is being sent

hugoroy:
I need to figure out where to get the variable to display the Cancel keybinding.

Also, maybe the "hint" could be hidden if the user has configured for less hints.

### DIFF
--- a/src/compose_message.cc
+++ b/src/compose_message.cc
@@ -371,7 +371,7 @@ namespace Astroid {
 
     if (cancel_send_during_delay) {
       LOG (error) << "cm: cancelled sending before message could be sent.";
-      message_send_status_msg = "sending message.. cancelled before sending.";
+      message_send_status_msg = "sending message... cancelled before sending.";
       message_send_status_warn = true;
       d_message_send_status ();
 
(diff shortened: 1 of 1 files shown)


gauteh:
Looks good, agree, user keybinding would be ideal (not yet included in
the other hints by the way). Also, I think the 'Ctrl+C can cancel' should
be shown when 'sending message..' as well.

Maybe both should only be shown when hint level > 1 like 'edit with
enter' in edit_message.

hugoroy:
> Also, I think the 'Ctrl+C can cancel' should
be shown when 'sending message..' as well. [...]
Maybe both should only be shown when hint level > 1 like 'edit with
enter' in edit_message.

Yes, it makes sense to have it there as well. I will add it and make it optional on the hint level like edit_message.

> Looks good, agree, user keybinding would be ideal (not yet included in
the other hints by the way).  

Ha, if it's not already implemented then I guess it makes sense to implement that overall in other places with another pull request.

gauteh:
Hugo Roy writes on mars 3, 2017 11:12:
>> Looks good, agree, user keybinding would be ideal (not yet included in
>> the other hints by the way).  
> 
> Ha, if it's not already implemented then I guess it makes sense to implement that overall in other places with another pull request.
> 

Yes - easier in edit_message than in compose_message (your case) since
in edit_message you have access to the keybindings already.

gauteh:
Thanks, merged. Sorry that it took some time.

PR #306, opened 2017-03-13
Multiple UI errors caused by missing shared files

jimcheetham:
This line should _extend_ the current definition of inst_shr, not replace it.

There are two main user symptoms of this: the "New Message" tab is completely
blank (I guess there is no HTML for it to render), and the "no-mail.png" icon
is missing, which sometimes makes the program fail on startup.


jimcheetham:
Closed in favor of #305 which covers the same problem

PR #308, opened 2017-03-15
Escaping . characters for CSS auto-tag-classes

mxmehl:
Extending the escaping mechanism from https://github.com/astroidmail/astroid/commit/0ff5fe1a70558c9f61001e9734e42d9df608f3d6 for people using periods in their notmuch tags.


gauteh:
Thanks, merged.

PR #314, opened 2017-04-06
compose: skip empty references/in-reply-to

TristanCacqueray:
This change avoid empty References/In-Reply-To header on new sent mail.

### DIFF
--- a/src/compose_message.cc
+++ b/src/compose_message.cc
@@ -83,14 +83,24 @@ namespace Astroid {
 
   void ComposeMessage::set_references (ustring _refs) {
     references = _refs;
-    g_mime_object_set_header (GMIME_OBJECT(message), "References",
-        references.c_str());
+    if (references.empty ())
+        g_mime_header_list_remove (
+            g_mime_object_get_header_list (GMIME_OBJECT(message)),
+            "References");
+    else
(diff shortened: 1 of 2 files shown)


gauteh:
Hi,

Thanks, I think that ideally this should remove the
In-Reply-To/References header if it already exists, since we might in
the future keep the ComposedMessage object around a litt longer (I don't
think this can happen at the current moment). So I think it would be
better to do g_mime_header_list_remove on the GMimeHeaderList (if it
exists).

This could be tested in test/test_composed_message in a similar
way as compose_test_body.

Regards, Gaute

Tristan de Cacqueray writes on april 6, 2017 4:24:

TristanCacqueray:
Hello,

Thanks for the feedback, I've updated the PR according to your suggestions, the test is failing without the src/compose_message.cc change.

Regards,
-Tristan

gauteh:
Thanks, merged (dc03dac). I did some final cleanups to make the code
adhere to coding style
(https://github.com/astroidmail/astroid/wiki/Style), and I added one
more `set_references()` call to the test to see what happens if you try
to remove the refrences header when it is not set from before.

- gaute

Tristan de Cacqueray writes on april 6, 2017 10:22:

Issue #316, opened 2017-04-10
Improve fast reading of mails


kirschner:
First of all, thank you very much for the huge improvements over the last months! 

I am wondering if there is a faster way to read my mails. At the moment I have to hit "return" to open an e-mail. After reading I close it with "ctrl+w". I have to switch to the next message with "tab", and start the same procedure again. 

With my mutt setup I am still much faster. I open an e-mail once with "return", and then I just press "j" to open the next unread message (either in the same thread or opening the next unread thread). 

Is there a way to achieve that with astroid? Or would that be something which could be enabled in future versions? My feeling is that this will significantly improve reading speed with astroid. 

Thanks a lot!

mxmehl:
Well, in the same thread you can always jump to the next unread mail by pressing `Tab`.

If it's about mails in another thread, this could be a bit tricky as there may be more than one thread_index with unread mails. So how could astroid know what the next unread thread is supposed to be? Maybe the last visited thread index from which the current thread_view has been opened?

TristanCacqueray:
It would be nice to have a new keybinding that combine next_unread and open_paned. Ideally that function should:

* go to next unread message in thread if pane is already open
* when thread is read, go to next thread with unread message
* when all thread are read, go to next saved search with unread thread

This should be similar to the Go to Next Unread Message of thunderbird.

rnav:
This happens to be one of the major issues for me at the moment. The ability to quickly scan through messages is very important when going through mailing lists, especially those with a high volume. In many ways, I miss the split pane view mutt provides.

gauteh:
rnav writes on april 23, 2017 17:39:
> This happens to be one of the major issues for me at the moment. The ability to quickly scan through messages is very important when going through mailing lists, especially those with a high volume. In many ways, I miss the split pane view mutt provides.

Have you tried the open_paned view? Does that provide some of that
functionality?

gauteh:
Please try #327

gauteh:
Fixed in #327.

rnav:
Getting back to this after a long time (sorry!), but open_paned view didn't help much. I have however setup keybindings for #327 and it does solve this particular issue, so Thanks!

PR #327, opened 2017-04-24
Thread-View: Archive and/or request action from originating thread-index

gauteh:
#316 and some other issue I can't remember, please try and comment on key choices (make sure new suggestions are consistent and free!)


gauteh:
There are now aliases in the main key namespace so that you do not need to press two keys to achieve any of these actions:
```
thread_view.archive_then_next -> thread_view.multi_next_thread.archive
etc..
```

gauteh:
Merged.

PR #336, opened 2017-05-05
thread-index: remove marked icon, move other icons to tag list

gauteh:
Goal is to make thread-index a bit better, hope to get rid of the icons on the left which are often left empty.

* Uses background color to indicate marked messages, thread-view must be updated to be consistent. Color is really crap at the moment. Suggestions appreciated.

### DIFF
--- a/History.txt
+++ b/History.txt
@@ -1,5 +1,7 @@
 ==
 
+* Move icons in thread-index to tag line
+* Allow user configured tag icons
 * Use color to indicate whether a message is marked in thread-index.
 
 == v0.10.2 / 2017-10-11
--- a/src/config.cc
+++ b/src/config.cc
@@ -181,7 +181,7 @@ namespace Astroid {
     default_config.put ("thread_index.cell.tags_upper_color", "#e5e5e5");
     default_config.put ("thread_index.cell.tags_lower_color", "#333333");
     default_config.put ("thread_index.cell.tags_alpha", "0.5");
-    default_config.put ("thread_index.cell.hidden_tags", "attachment,flagged,unread");
+    default_config.put ("thread_index.cell.hidden_tags", "unread");
 
 
     /* editor */
--- a/src/modes/thread_index/thread_index_list_cell_renderer.cc
+++ b/src/modes/thread_index/thread_index_list_cell_renderer.cc
@@ -59,6 +59,15 @@ namespace Astroid {
     subject_color_selected = ti.get<string> ("subject_color_selected");
     background_color_selected = ti.get<string> ("background_color_selected");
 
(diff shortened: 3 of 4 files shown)


gauteh:
Thanks, that looks much better. (btw didn't realize you had made the changes yourself somehow before after I pushed the new commit)

gauteh:
Check out latest: icons have been moved to tag line and you can now configure your own tag to image maps using e.g.:
```
            "tag_images" : {
              "bra" : "/usr/share/icons/Humanity/status/48/sunny.svg"
            }
```

under `thread_index.cell` in the config.

2017-05-18-115047_761x800_scrot

mxmehl:
Sweet changes, but I'm not really happy with moving the flagged icon to the tags. This way, I will often overlook them.

Would it be possible to have a slightly different background for flagged messages? Or some other way to recognize them at first sight?

Regarding the attachment icon, the current solution is fine for me. I often don't specifically look for attachments so it's not that critical whether I see it at first sight when skimming over a long list of threads.

gauteh:
Good point about the flagged icon. Might be too much or confusing with marked to use a backgronud color, was thinking about underlining, but i think that gets messy. suggestions welcome.

gauteh:
Added some padding before the date field.

mxmehl:
It may be my unnatural sense of order but the attachment icons disturb the beautiful vertical tab line ;)

http://i.imgur.com/ry2nqeU.png

Can we move the attachment/flagged icons behind the tags? Furthermore, the attachment icon is normally less important than all the tags.

gauteh:
Max writes on mai 18, 2017 23:02:
> It may be my unnatural sense of order but the attachment icons disturb the beautiful vertical tab line ;)
> 
> http://i.imgur.com/ry2nqeU.png
> 
> Can we move the attachment/flagged icons behind the tags? Furthermore, the attachment icon is normally less important than all the tags.

Indeed, that doesn't look very good. I usually have too many tags, so I
wouldn't see the icons in that case..

mxmehl:
I don't know how to proceed here. The current space-saving behaviour contradicts my wish for a clear vertical line in front of the tags, and your wish to have the flagged/attachment icons in front of the taglist...

mxmehl:
Can you push the marked messages to master while we think about how to solve the flagged/attachment conflict?

gauteh:
Max writes on september 3, 2017 14:52:
> Can you push the marked messages to master while we think about how to solve the flagged/attachment conflict?

Yes, but need some good colors for marked messages. Suggestions?

gauteh:
Branch copy here: https://github.com/gauteh/astroid/tree/ti-dabblings

PR #341, opened 2017-05-09
move the query column to the right. My queries are over 600 character…

### LINKED ISSUE #340: saved search column scrolls off into the distance..
My saved searches are very very long (over 600 characters). That's why they are saved.. too long to make up ad-hoc every time.

This means that the unread/total column is never seen, it's way off into the right.

I vote to put the unread/total before the query. Other options would be to have the query column hideable or scrollable within it's own fixed-width.

yeled:
…s long sometimes. This scrolls on forever so I can't see the unread/total columns

kind-of fixes #340

### DIFF
--- a/src/modes/saved_searches.cc
+++ b/src/modes/saved_searches.cc
@@ -36,9 +36,9 @@ namespace Astroid {
       pcolumn->add_attribute (renderer_text->property_markup(), m_columns.m_col_name);
     }
 
-    tv.append_column ("Query", m_columns.m_col_query);
     tv.append_column ("Unread messages", m_columns.m_col_unread_messages);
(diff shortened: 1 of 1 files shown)


gauteh:
Charlie Allom writes on mai 9, 2017 13:59:
> …s long sometimes. This scrolls on forever so I can't see the unread/total columns
> 
> kind-of fixes #340
> You can view, comment on, or merge this pull request online at:

Thanks, I'm positive. Perhaps you could provide a screenshot?

yeled:
image

I hope this doesn't have too much incriminating evidence :)

gauteh:
Thanks, merged!

Issue #354, opened 2017-05-25
Build error: 'sh: 1: sassc: not found' though it is in $PATH


airways:
After getting past other build errors on Ubuntu 14.04 including installing libsass and sassc from source and your fix for #353, I am now having trouble with the build failing to find sassc:

```
iraway@OCULAR:~/workspace/astroid$ scons --disable-libsass
scons: Reading SConscript files ...
building version v0.9.1-13-g2240ef5c (git)..
debug flag enabled: True
Checking for pkg-config... (cached) yes
Checking for gtkmm-3.0 >= 3.10... (cached) yes
Checking for glibmm-2.4... (cached) yes
Checking for gmime-2.6 >= 2.6.18... (cached) yes
Checking for webkitgtk-3.0... (cached) yes
Checking for vte-2.91... (cached) no
warning: vte3 not found: disabling built-in terminal.
warning: built-in terminal disabled.
warning: libsass is disabled, will generate SCSS at build time using: 'sassc'..
Checking for gobject-introspection-1.0... (cached) yes
Checking for libpeas-1.0... (cached) yes
Checking for C library notmuch... (cached) yes
Checking for C++ library boost_filesystem... (cached) yes
Checking for C++ library boost_system... (cached) yes
Checking for C++ library boost_program_options... (cached) yes
Checking for C++ library boost_log_setup... (cached) yes
Checking for C++ library boost_log... (cached) yes
Checking for C++ library boost_thread... (cached) yes
Checking for C++ library boost_date_time... (cached) yes
writing src/build_config.hh..

    debug   ..:  True
    release ..:  git
    version ..:  v0.9.1-13-g2240ef5c
    profile ..:  False
    libsass ..:  False
    scss .....:  sassc ( use: True )
    plugins ..:  True
    terminal .:  False
    embedded .:  True
    prefix ...:  /usr/local

scons: done reading SConscript files.
scons: Building targets ...
compiling ==> src/astroid.cc
compiling ==> src/config.cc
compiling ==> src/main_window.cc
compiling ==> src/poll.cc
compiling ==> src/modes/edit_message.cc
compiling ==> src/modes/forward_message.cc
compiling ==> src/modes/help_mode.cc
compiling ==> src/modes/keybindings.cc
compiling ==> src/modes/log_view.cc
compiling ==> src/modes/mode.cc
compiling ==> src/modes/paned_mode.cc
compiling ==> src/modes/raw_message.cc
compiling ==> src/modes/reply_message.cc
[...]

PR #355, opened 2017-05-25
Adding note about --scss-compiler to libsass not found error

airways:
Added to libsass not found message to prevent others from running into issue in #354.


gauteh:
Isaac Raway writes on mai 25, 2017 23:48:
> Added to libsass not found message to prevent others from running into issue in #354.

Thanks, merged. Changed should to can since this dependes on your
environment. If scss is in non-system locatablable place you need to use
--propagate-environment to get all custom PATH and stuff.

PR #366, opened 2017-07-15
Cb tag jump

gauteh:
* Colors for tag-editor
* Better jump and delete between tags when editing

http://imgur.com/a/hw0qB

### DIFF
--- a/src/command_bar.cc
+++ b/src/command_bar.cc
@@ -294,6 +294,52 @@ namespace Astroid {
         }
         break;
 
+      case GDK_KEY_Left:
+        {
+          LOG (debug) << "cb: left";
+          if (mode == Tag || mode == DiffTag) {
+            /* jump to previous tag separator */
+            if (entry.get_position () == 0) return true;
+
+            refptr<TagCompletion> t = refptr<TagCompletion>::cast_dynamic (current_completion);
+
+            ustring_sz pos;
+            t->get_partial_tag (pos);
+
+            if ((int) pos == entry.get_position ()) {
+              /* at separator, search backwards for next tag and jump to its beginning */
+
+              ustring txt = entry.get_text ();
+
+              pos = (pos > 1) ? pos - 1 : pos;
+
+              while (pos > 1) {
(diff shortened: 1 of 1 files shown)


mxmehl:
The colours are very nice. The only thing I don't like here is that as soon as the tags are modified (add or delete a character), the font colour changes to white instead of black. This makes it hard to read.

The jump between the tags (using the left arrow key?) doesn't work for me.

gauteh:
Max writes on juli 17, 2017 19:03:
> The colours are very nice. The only thing I don't like here is that as soon as the tags are modified (add or delete a character), the font colour changes to white instead of black. This makes it hard to read.
> 
> The jump between the tags (using the left arrow key?) doesn't work for me.

Yeah, there's a bug about the font color still (and the color of the 
last tag). Not sure about the jumping yet, I get to the right one when I 
use Alt-Left. Thinking about moving it to Ctrl+h. Also, I'd like to 
select the whole tag when moving left so that the whole tag can be 
easily removed with just Backspace. Making it easier to remove tags.

gauteh:
Gaute Hope writes on juli 17, 2017 19:35:
> Max writes on juli 17, 2017 19:03:
>> The colours are very nice. The only thing I don't like here is that as soon as the tags are modified (add or delete a character), the font colour changes to white instead of black. This makes it hard to read.
>> 
>> The jump between the tags (using the left arrow key?) doesn't work for me.
> 
> Yeah, there's a bug about the font color still (and the color of the 
> last tag). Not sure about the jumping yet, I get to the right one when I 
> use Alt-Left. Thinking about moving it to Ctrl+h. Also, I'd like to 
> select the whole tag when moving left so that the whole tag can be 
> easily removed with just Backspace. Making it easier to remove tags.
> 
> 

font color should be fixed, as well as colors for unicode chars

gauteh:
Pushed colorful tags to master, tag-jump not yet.

gauteh:
branch copy here: https://github.com/gauteh/astroid/tree/cb-tag-jump

PR #368, opened 2017-07-22
cb: do not use ',' to delimit tags, only space

gauteh:
Skip the ',' as delimiter for tags. This makes it quicker to navigate and edit tags. Tags with spaces in them are not supported, but I dont think they were in the first place either. Opinions?

### DIFF
--- a/History.txt
+++ b/History.txt
@@ -2,6 +2,10 @@
 
 * Attach messages by mids with 'A' in new emails.
 
+* Colorful tags in tag editor and in searching
+
+* Tags in the tag editor is now separated by spaces (comma still works)
+
 == v0.9.1 / 2017-04-30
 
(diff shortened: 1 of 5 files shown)


yeled:
sgtm 👍

mxmehl:
Will there be backwards compatibility? What will happen if I keep using `,` to separate tags?

gauteh:
At the moment the , will be considered part of the tag or a tag in itself.

You can also see this from the backgroundcolor including the ,

It wouldnt be a big deal i think to also support ,

søn. 23. jul. 2017 kl. 19.10 skrev Max <notifications@github.com>:

gauteh:
Now splitting on both.

gauteh:
Merged.

Issue #371, opened 2017-07-25
Add support for a signature separator, typically '-- \n'


mrvdb:
If I want to use a signature separator I have to include it in my signature file at the moment, which is arguably the wrong place.

If there was support for a signature separator, this would also be helpful for:
- perhaps help resolve #234 
- in thread view, anything after the signature separator could be rendered different from the main body, or left out even

gauteh:
Marcel van der Boom writes on juli 25, 2017 12:01:
> If I want to use a signature separator I have to include it in my signature file at the moment, which is arguably the wrong place.
> 
> If there was support for a signature separator, this would also be helpful for:
> - perhaps help resolve #234 
> - in thread view, anything after the signature separator could be rendered different from the main body, or left out even
> 

Hi,

We used to have the standard signature separator. However, I think it is 
more flexible this way. People use all kinds of signatures, and 
because of this we can never rely on parsing an e-mail backwards to 
really find the signature, since this string could easily appear several 
times in the signature.

To fix #234 we should use the Xapian metadata properties which are available in notmuch now and add them to the draft. Then we can store state information there untill the draft is sent.

I agree with the final point, also this will be kind of hard to parse - 
but if we can come up with a consistent algorithm that detects what part 
is signature with a very low degree of false positives I think it should 
be included.



Regards, Gaute

mrvdb:
What flexibility do we lose if there was, say, a config item `accounts.myaccount.signature_separator` ? 
(defaulting to "" I guess)

gauteh:
Marcel van der Boom writes on juli 25, 2017 12:33:
> What flexibility do we lose if there was, say, a config item `accounts.myaccount.signature_separator` ? 
> (defaulting to "" I guess) 

Would it not be easier to just edit the signature file in that case?

mrvdb:
If you're using just one MUA, yes, perhaps.

gauteh:
Marcel van der Boom writes on juli 25, 2017 12:37:
> If you're using just one MUA, yes, perhaps.

How does mutt and others do it? If the '--  \n' is prepended by default 
I'd be more positive to adding it as config option. In that case it 
should be on by default.

mrvdb:
my gnus and thunderbird config default to having it prepended to the signature file.  Thunderbird can be made to not do that, but havent looked at gnus yet (I'm sure it's possible)

gauteh:
Marcel van der Boom writes on juli 25, 2017 12:41:
[...]

PR #372, opened 2017-07-25
Introduce signature_separate config option

mrvdb:
When true, the signature part of the message will be separated from
the message by '-- \n'

Related issue #371

### DIFF
--- a/src/account_manager.cc
+++ b/src/account_manager.cc
@@ -39,6 +39,7 @@ namespace Astroid {
 
       a->save_drafts_to = kv.second.get<string> ("save_drafts_to");
 
+      a->signature_separate = kv.second.get<bool> ("signature_separate");
       a->signature_file = kv.second.get<string> ("signature_file");
       a->signature_default_on = kv.second.get<bool> ("signature_default_on");
       a->signature_attach     = kv.second.get<bool> ("signature_attach");
--- a/History.txt
+++ b/History.txt
@@ -6,6 +6,9 @@
 
 * Tags in the tag editor is now separated by spaces (comma still works)
 
+* configuration option signature_separate determines if signatures
+  will be separated from the message by a line containing '-- '.
+
 == v0.9.1 / 2017-04-30
 
 * Support old vte3.
--- a/src/account_manager.hh
+++ b/src/account_manager.hh
@@ -24,6 +24,7 @@ namespace Astroid {
       std::vector<ustring> additional_sent_tags;
       ustring save_drafts_to;
 
+      bool      signature_separate = false;
       bfs::path signature_file;
       bool      signature_default_on;
       bool      signature_attach;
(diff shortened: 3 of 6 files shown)


src/account_manager.cc
```diff
@@ -39,6 +39,7 @@ namespace Astroid {
 
       a->save_drafts_to = kv.second.get<string> ("save_drafts_to");
 
+      a->signature_separate = kv.second.get<bool> ("signature_separate");
```
gauteh: on second thought this should probably be "signature_separator" ?

gauteh:
Looks good: we need to support seamless config-update on this one, probably have to bump config version as well. Have a look here for earlier examples: https://github.com/astroidmail/astroid/blob/master/src/config.cc#L431

We can also probably start to remove the old migrations - I don't think we need to support config files that are more than 2 years old.

src/account_manager.cc
mrvdb: I had it like that first (as a string config option) which specified the separator, if it is a boolean 'signature_separate' is better, no?

gauteh: ok! sounds good.

gauteh:
Also, you have to add the default setting to the example config in config.cc.

mrvdb:
Something like this you mean?

```
diff --git a/src/config.cc b/src/config.cc
index 782dac3..71401fd 100644
--- a/src/config.cc
+++ b/src/config.cc
@@ -146,6 +146,7 @@ namespace Astroid {
       default_config.put ("accounts.charlie.save_drafts_to",
           "/home/root/Mail/drafts/");
 
+      default_config.put ("accounts.charlie.signature_separate", false);
       default_config.put ("accounts.charlie.signature_file", "");
       default_config.put ("accounts.charlie.signature_default_on", true);
       default_config.put ("accounts.charlie.signature_attach", false);
@@ -439,6 +440,18 @@ namespace Astroid {
             config.put (key.c_str (), false);
           }
         }
+
+       if (version < 8) {
+         try {
+           
+           ustring ss = kv.second.get<bool> ("signature_separate");
+           
+         } catch (const boost::property_tree::ptree_bad_path &ex) {
+           
+           ustring key = ustring::compose ("accounts.%1.signature_separate", kv.first);
+           config.put (key.c_str (), false);
+          }
+       }
       }
 
       changed = true;
```

gauteh:
Marcel van der Boom writes on juli 25, 2017 14:53:
> Something like this you mean?
> 
> ```
> +           config.put (key.c_str (), false);
> ```

Should this be the same as the default? This will 
cause a bit of a hassle for users, but we'll have to write it in the 
History.txt / release announcement. Also, it's not the end of the world 
if there is a few double '--'s.

Please test if upgrading the config works without hassle (the option 
should be added without error if it is not present).

Config version is set in config.hh if I remember correctly.

- gaute

gauteh:
Otherwise the config stuff looks good :)

mrvdb:
I think i've addressed all hints and tips (thanks!) and from my point of view, this can be merged.

gauteh:
Thanks, merged! I fixed up the commit messages and tweaked the History a bit.

Issue #381, opened 2017-08-15
'c' on message does not include Cc (and probably Bcc) in recipients of new message


gauteh:
should include all recipients.

mrvdb:
To be clear, is this the desired behaviour?

 - copy `Cc:` into `Cc:` of new message
 - copy `Bcc:` into `Bcc:` of new message

If message is from own account:
  - copy `From:` into `From:` of new message
  - copy `To:` into `To:` of new message 

else:
  - copy `From:`  to `To:` in new message
  - use active account `From:` for new message

gauteh:
Marcel van der Boom writes on august 15, 2017 12:51:
> To be clear, is this the desired behaviour?
> 
>  - copy `Cc:` into `Cc:` of new message
>  - copy `Bcc:` into `Bcc:` of new message

Yes, but only if is from own account.

> If message is from own account:
>   - copy `From:` into `From:` of new message
>   - copy `To:` into `To:` of new message 
> 
> else:
>   - copy `From:`  to `To:` in new message
>   - use active account `From:` for new message 

The account matching the 'To:' in the old message.

mrvdb:
Resolved with 8a3b443886442009c924807c031421c4b8c9b96b

PR #382, opened 2017-08-15
Improve 'c' (compose to sender) command

mrvdb:
Initial resolve for issue #381  but probably needs a bit more, please
review.

- clarify help text for command
- support cc and bcc in recipient when composing to self
  (bcc is probably not that useful here)
- adjust variable naming to clarify code

Added `EditMessage` with additional signature, probably a bit
overkill and could use just the one.

### DIFF
--- a/src/modes/edit_message.cc
+++ b/src/modes/edit_message.cc
@@ -34,12 +34,13 @@ using namespace boost::filesystem;
 namespace Astroid {
   int EditMessage::edit_id = 0;
 
-  EditMessage::EditMessage (MainWindow * mw, ustring _to, ustring _from) :
+  EditMessage::EditMessage (MainWindow * mw, ustring _to, ustring _from, ustring _cc, ustring _bcc) :
     EditMessage (mw, false) { // {{{
 
     in_read = false;
-    to = _to;
-
+    to  = _to;
+    cc  = _cc;
+    bcc = _bcc;
     if (!_from.empty ()) {
       set_from (Address (_from));
     }
--- a/src/modes/edit_message.hh
+++ b/src/modes/edit_message.hh
@@ -31,7 +31,7 @@ namespace Astroid {
 
     public:
       EditMessage (MainWindow *, bool edit_when_ready = true);
-      EditMessage (MainWindow *, ustring to, ustring from = "");
+      EditMessage (MainWindow *, ustring to, ustring from = "", ustring cc = "", ustring bcc = "");
       EditMessage (MainWindow *, refptr<Message> _msg);
       ~EditMessage ();
 
--- a/src/modes/thread_view/thread_view.cc
+++ b/src/modes/thread_view/thread_view.cc
@@ -2440,22 +2440,26 @@ namespace Astroid {
         });
 
     keys.register_key ("c", "thread_view.compose_to_sender",
-        "Compose a new message to the sender of the message (or receiver if sender is you)",
+        "Compose a new message to the sender of the message (or all recipients if sender is you)",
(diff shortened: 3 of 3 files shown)


src/modes/thread_view/thread_view.cc
```diff
@@ -2440,34 +2440,36 @@ namespace Astroid {
...
           if (!edit_mode) {
-            Address _to = focused_message->sender;
-            AddressList to;
-
-            if (astroid->accounts->is_me (_to)) {
-              to = AddressList(focused_message->to ());
-              Address from = _to;
-              main_window->add_mode (new EditMessage (main_window, to.str (), from.full_address ()));
+            Address sender = focused_message->sender;
+	    Address from;
```
gauteh: fix indentation

src/modes/thread_view/thread_view.cc
```diff
@@ -2440,34 +2440,36 @@ namespace Astroid {
...
+            AddressList to, cc, bcc;
+
+	    /* Send to original sender if message is not from own account,
+	       otherwise use all recipients as in the original */
+            if (astroid->accounts->is_me (sender)) {
+              from = sender;
+              to   = AddressList(focused_message->to  ());
+              cc   = AddressList(focused_message->cc  ());
+              /* Q: can Bcc ever appear in a message received? */
```
gauteh: No, but this is a message sent - can remove comment

src/modes/edit_message.cc
```diff
@@ -51,6 +51,26 @@ namespace Astroid {
     edit_when_ready ();
   }
 
+  EditMessage::EditMessage (MainWindow * mw, ustring _to, ustring _cc, ustring _bcc, ustring _from) :
```
gauteh: I think we should rather extend the signature with the from = "" argument, feel free to re-org arguments and make sure to grep for other places it is used.

mrvdb: Agreed

src/modes/thread_view/thread_view.cc
mrvdb: Yes, what I meant was, Bcc will never appear in a message that you want to compose-to-sender, except when it is a Draft, unless I am missing something. 
So copying it is a no-op most of the cases.

src/modes/thread_view/thread_view.cc
mrvdb: You're right, I need to add the style into my Emacs config

src/modes/thread_view/thread_view.cc
mrvdb: Ah, bcc stays in place when "sending" from astroid indeed. Never mind.

gauteh:
Yep, msmtp removes them (unless configured otherwise), we should update the
wiki to say users *must* configure their MTA to strip bcc headers.


tir. 15. aug. 2017 kl. 16.39 skrev Marcel van der Boom <
notifications@github.com>:

src/modes/edit_message.hh
```diff
@@ -31,7 +31,7 @@ namespace Astroid {
 
     public:
       EditMessage (MainWindow *, bool edit_when_ready = true);
-      EditMessage (MainWindow *, ustring to, ustring from = "");
+      EditMessage (MainWindow *, ustring to, ustring from = "", ustring cc="", ustring bcc="");
```
gauteh: make sure to use same formatting as _from with space either side of '='.

mrvdb:
Fixed spacing, should be good to go now.

gauteh:
Thanks, merged!

PR #401, opened 2017-09-17
Add log-level as command line parameter

mrvdb:
Adds support for setting log-level to command line parameters.

### DIFF
--- a/src/astroid.cc
+++ b/src/astroid.cc
@@ -10,6 +10,9 @@
 # include <boost/program_options.hpp>
 # include <boost/filesystem.hpp>
 
+/* for map_list_of */
+# include <boost/assign/list_of.hpp>
+
 /* log */
 # include <boost/log/core.hpp>
 # include <boost/log/utility/setup/file.hpp>
@@ -22,6 +25,7 @@
 # include <boost/log/utility/setup/formatter_parser.hpp>
 # include <boost/date_time/posix_time/posix_time_types.hpp>
 # include <boost/log/expressions.hpp>
+# include <boost/log/trivial.hpp>
 # include <boost/log/support/date_time.hpp>
 
 # include "astroid.hh"
@@ -119,6 +123,7 @@ namespace Astroid {
       ( "log,l", po::value<ustring>(), "log to file")
       ( "append-log,a", "append to log file")
       ( "disable-log", "disable logging")
+      ( "log-level", po::value<ustring>(), "set logging level (trace, debug, info, warning, error, fatal)")
       ( "start-polling", "indicate that external polling (external notmuch db R/W operations) starts")
       ( "stop-polling", "indicate that external polling stops")
       ( "refresh", po::value<unsigned long>(), "refresh threads changed since lastmod")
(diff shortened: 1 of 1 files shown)


src/astroid.cc
```diff
@@ -158,6 +163,16 @@ namespace Astroid {
...
+    } else if (vm.count ("log-level")) {
+      ustring llevel = vm["log-level"].as<ustring> ();
+
+      /* Not very elegant map for string to enum */
+      std::map<std::string, logging::trivial::severity_level> sevmap = boost::assign::map_list_of
+        ("trace", logging::trivial::trace)("debug", logging::trivial::debug)
+        ("info", logging::trivial::info)("warning", logging::trivial::warning)
+        ("error", logging::trivial::error)("fatal", logging::trivial::fatal);
+
+       logging::core::get()->set_filter (logging::trivial::severity >= sevmap[llevel]);
```
gauteh: What if llevel is not one of the legal options?

src/astroid.cc
```diff
@@ -158,6 +163,16 @@ namespace Astroid {
...
       LOG (warn) << "disabling log";
       logging::core::get()->set_logging_enabled (false);
+    } else if (vm.count ("log-level")) {
+      ustring llevel = vm["log-level"].as<ustring> ();
+
+      /* Not very elegant map for string to enum */
+      std::map<std::string, logging::trivial::severity_level> sevmap = boost::assign::map_list_of
+        ("trace", logging::trivial::trace)("debug", logging::trivial::debug)
+        ("info", logging::trivial::info)("warning", logging::trivial::warning)
+        ("error", logging::trivial::error)("fatal", logging::trivial::fatal);
```
gauteh: Is it possible to re-write this to only use standard C++14 or later? It's best to use Boost as little as possible, then we avoid the extra `include` as well.

src/astroid.cc
mrvdb: it will be silently ignored

src/astroid.cc
mrvdb: probably possible, this was just the easiest I could come up with.

out of curiosity, why use boost as little as possible, isn't the goal of a library to be used. Is there a severe penalty involved?

gauteh:
Marcel van der Boom writes on september 18, 2017 13:57:
> mrvdb commented on this pull request.
> 
> 
> 
>> @@ -158,6 +163,16 @@ namespace Astroid {
>      if (vm.count ("disable-log")) {
>        LOG (warn) << "disabling log";
>        logging::core::get()->set_logging_enabled (false);
> +    } else if (vm.count ("log-level")) {
> +      ustring llevel = vm["log-level"].as<ustring> ();
> +
> +      /* Not very elegant map for string to enum */
> +      std::map<std::string, logging::trivial::severity_level> sevmap = boost::assign::map_list_of
> +        ("trace", logging::trivial::trace)("debug", logging::trivial::debug)
> +        ("info", logging::trivial::info)("warning", logging::trivial::warning)
> +        ("error", logging::trivial::error)("fatal", logging::trivial::fatal);
> 
> probably possible, this was just the easiest I could come up with.
> 
> out of curiosity, why use boost as little as possible, isn't the goal of a library to be used. Is there a severe penalty involved?

boost is pretty modular, so the less we use the less we need it. 

Ideally, if something is available in the standard C++ library we use 
that. An auxilliary library is unlikely to beat that in terms of code 
size, performance, clarity, maintainability, and stability. If it is in 
standard library it is easier for the next developer who reads your code 
to understand it. Docs and behavior would correspond to standard C++ and 
not the quirks of a specific library.

If we somewhen manage to get rid of a boost dependency that would make 
astroid significantly less complex, the less we use it, the easier that is.

src/astroid.cc
gauteh: ok, maybe put in a comment

gauteh:
Thanks, merged!

PR #408, opened 2017-09-25
Webkit2js

gauteh:
(replaces #402 )


Ramblurr:
Do you want to create a legacy branch for fixes/updates to the webkit1 version? As if this goes into master, you can't release until we finish the JS implementation. I've had some time recently, but I'm not sure how much longer it will take or if it will be done sooner rather than later.

gauteh:
Yes, probably - but we're starting to get quite a bit of trouble with webkit1.

gauteh:
I feel like message loading and display is a bit slow. Is this just an artifact of the inspector window and debugging or do you experience the same?

gauteh:
Now closing in favor of the merged #455. Thanks for the massive effort!

PR #412, opened 2017-10-02
Move to the Meson build system

gauteh:
[Meson](http://mesonbuild.com/) is modern, way faster, and is not scons.

```
$ meson build
$ ninja -C build
$ build/astroid # start astroid
```

`examples/astorid` has also been updated (assuming `build` is used for build directory).

TODO:
- [X] Use sassc when no libsass
- [X] Install
- [X] Tests
- [X] Check for notmuch function
- [] Enable profiling if not supported by meson already
- [X] Remove scons

Meson availability:

- [X] Arch Linux
- [X] Debian / Ubuntu
- [X] Fedora
- [X] Homebrew

Documentation updates to do:
- [X] README
- [ ] compile_commands
- [ ] Version number in meson.build
- [ ] All references to scons


Ramblurr:
Just curious, but have you considered cmake?

gauteh:
Casey Link writes on oktober 2, 2017 13:18:
> Just curious, but have you considered cmake?

I've found it horrendous in the past.

c-alpha:
@gauteh wrote: 
> Casey Link writes on oktober 2, 2017 13:18:
>> Just curious, but have you considered cmake?
> 
> I've found it horrendous in the past.

I have used cmake in industrial projects. Many big corps use it, and quite a couple of OSS projects use it, too. Looking at its build file syntax, you will certainly get a scent of its age (it started 1999). Meson looks very similar in both, where it sits in terms of abstraction level, and build file syntax. Back in the early days (I first laid hands on cmake ca. 2002) cmake's syntax was not too dissimilar to meson's. As people used - and therefore extended it - more and more, the syntax grew more, let's say "baroque". I dare to predict that the same will happen to meson over time.

Besides all it's quirks and historically grown build file syntax, one thing I can say about cmake is that it is both, fast  and rock solid. Once you've gotten it to do what you want, it will dependably do it for you for the foreseeable future. It buys you peace of mind.

One reason - IMO - for this is that cmake has **NO** external dependencies. It ships with the tried & tested versions of everything it needs (curl, zlib, etc.), so you can be sure that the same version of cmake will do _exactly_ the same thing, even when resurrected from a tape archive three major OS releases later. To me this reproducibility is a key property of a stable and dependable build system.

Meson on the other hand relies on python3 and ninja as external tools. If one of them gets upgraded, it may break your builds. And that will be really hard to debug.

Also, python2 is still very popular. Hence, it will mean for many (i.e. including my humble self) that they will need to have two versions of python installed in parallel. Not to say it can't be done, but its another paint point, and another source for breakage.

In summary, I must thus concede that I view the choice of meson to be somewhat unfortunate. Yes, cmake has its quirks. Like every good tool does. An - IMO - important aspect is however that so many people still using cmake very likely means that they all have found ways to live every day with its quirks, which suggests that cmake's quirks can't be that bad after all.

**Caveat:** This is not so say that I demand cmake. But I would highly welcome if a build system could be chosen that does not force me to install python3.

c-alpha:
Update on parallel use of python 2 and 3: you can use either of: conda, virtualenv, or the new py launcher. But in all cases I have to manually decide (i.e. either activate the right environment, or pass the right command line option) which python version to use for each script invocation.

Since I will soon forget what version each script needed, this will likely boil down to an epical size pile of `alias` definitions in my `bashrc`. Just saying...

gauteh:
Hi, thanks for the comments! I agree - I see some of the same things in 
meson. Just a quick reply regarding python version; by now python3 is 
available on any system that astroid works on. It is probably also the 
default version there by now. SCons on the other hand is mostly python2, 
as you noticed when you tried to build on Mac OS and had to essentially 
downgrade your python and scons. 

It is probably possible to achieve the same functionality with CMake as 
meson, but if you take a look at the meson.build you will see that a lot 
of stuff is done automatically now. Both meson and ninja are used on big 
projects (e.g. chromium).

c-alpha:
I couldn't hold back, and have created a branch in my fork that uses cmake:
https://github.com/c-alpha/astroid/tree/cmake

It builds a debug version of astroid, which runs fine as far as I can tell.
From the commit message:
> TODO: release build
> TODO: check other platforms (enable plugins, internal editor)
> TODO: tests

gauteh:
That does look pretty good. also missing check  for notmuch function,
profile build, do you build the css file using sassc when no libsass?, test
setup.

tor. 5. okt. 2017 kl. 03:41 skrev c-alpha <notifications@github.com>:

gauteh:
And GIR generation.

gauteh:
Casey Link writes on oktober 5, 2017 11:50:
> wrt libsass, with the new js stuff coming soon we might as well generate the `css` file from the `scss` when the javascript is compiled (I already am locally). this would let astroid drop scss/sass as a build and runtime dep.
> 

Yes, the reason for keeping sass is that it enables users to theme the 
thread-view more easily by just changing scss variables (font, colors).

c-alpha:
How did you hack into my todo list?
;-))

  --alex

-- 
www.condition-alpha.com / @c_alpha

c-alpha:
I have enabled issues on my cmake fork (https://github.com/c-alpha/astroid/tree/cmake).

Current issues: https://github.com/c-alpha/astroid/issues

If anything else comes to mind that should be added, or something should be done first, please don't hesitate to create a new issue, or add a comment to an existing one.

c-alpha:
> [...]
> also missing check  for notmuch function,
> [...]

@gauteh, couldn't that be easier checked via a notmuch's version number? It seems this API was only introduced with 0.25.1?

Ideally, you could you answer as a new comment to c-alpha/astroid#3 please? Thanks!

gauteh:
I've tested your CMake branch a little; a few differences with Meson:

* Both CMake and Meson can use Ninja which is way faster than Autotools/Make
* CMake requires you to create the build directory first
* Running make/ninja inside the build directory re-generates the makefiles/build.ninja files automatically for Meson, but not for CMake
* Meson tests support the ninja test system, I don't know if CMake does this
* GIR (https://github.com/c-alpha/astroid/issues/4) is very easy with Meson, but there might be a CMake module for this already.
* It seems that e.g. [Nautilus](https://github.com/GNOME/nautilus) and many other [GNOME packages](https://github.com/search?q=org%3AGNOME+meson.build&type=Code) are moving to Meson - but some keep dual build systems. On that note I wouldn't have a problem with having both Meson and CMake, though having to support two build systems is not ideal. Just because of the availability and stability of CMake.

gauteh:
Unless we find a workaround this blocks on Mac OS X: https://github.com/mesonbuild/meson/issues/2436

gauteh:
I have merged the addition of the meson system, but I have not yet removed scons (https://github.com/gauteh/astroid/tree/remove-scons). @c-alpha: If you get a fully functional CMake system up and going we can have both, and I will remove SCons, but I will leave scons for a while yet on master at least.

gauteh:
@c-alpha: Any progress on CMake? I'd really like to include this, SCons is starting to move to scons3 everywhere and its a pain. Meson is not yet really stable.

c-alpha:
Hello @gauteh. Sorry for not having responded earlier. I am pretty tied up this week with travel & meetings. Its not forgotten, and I will tend to it next week.

I think your idea of having both, and then potentially taking a pick is a good one. 

Thanks for bearing with me!

gauteh:
No worries :)

c-alpha writes on oktober 9, 2017 17:38:
> Hello @gauteh. Sorry for not having responded earlier. I am pretty tied up this week with travel & meetings. Its not forgotten, and I will tend to it next week.
> 
> I think your idea of having both, and then potentially taking a pick is a good one. 
> 
> Thanks for bearing with me!
>

gauteh:
Just a list of my current meson issues: https://github.com/mesonbuild/meson/issues/created_by/gauteh

gauteh:
Jonathan Lambrechts writes on oktober 17, 2017 15:23:
> Hello,
> as discussed on https://aur.archlinux.org/packages/astroid/, meson fails to find boost if the language of the current locale is not english see this mesonbuild bug report https://github.com/mesonbuild/meson/issues/2501
> 

Right, I've updated the PKGBUILD's now.

gauteh:
@c-alpha: I'm rooting for the CMake build! Meson is definetely causing too many problems!

c-alpha:
On 2017-11-02, at 15:52 , Gaute Hope <notifications@github.com> wrote:

> @c-alpha: I'm rooting for the CMake build! Meson is definetely causing too many problems!
> [...]

Thanks for letting me know.

What do I take from "rooting"?

gauteh:
c-alpha writes on november 2, 2017 22:57:
> On 2017-11-02, at 15:52 , Gaute Hope <notifications@github.com> wrote:
> 
>> @c-alpha: I'm rooting for the CMake build! Meson is definetely causing too many problems!
>> [...]
> 
> Thanks for letting me know.
> 
> What do I take from "rooting"?

Cheering :)

PR #413, opened 2017-10-03
Fix macOS Build

c-alpha:
There were two issues:

1. Fix in src/utils/compiler.h

   The build failed when compiling
   src/utils/gmime/gmime-filter-html-bq.c. It complains about a missing
   ')' after 'clang' in line 5 of compiler.h. This is because
   clang::fallthrough is a scoped attribute, but the used macro
   __has_attribute will only accept single identifiers under
   clang. Hence, its sibling __has_cpp_attribute must be used. Then,
   when using the LLVM clang that ships with Xcode, the
   __has_cpp_attribute is not defined at all when compiling
   gmime-filter-html-bq.c. This is because it is a C file, and the LLVM
   clang that ships with Xcode hides the C++ features (including the
   feature checking macros) when operating in C mode. Thus we must
   safeguard against the __has_* macros not being defined. For further
   details, see
   https://isocpp.org/std/standing-documents/sd-6-sg10-feature-test-recommendations
   and
   https://clang.llvm.org/docs/LanguageExtensions.html#feature-checking-macros

2. Fix in SConstruct

   When linking, there were tons of undefined symbol errors against
   stuff from boost log. As pointed out in this stackoverflow answer,
   and as oft:
   https://stackoverflow.com/questions/38121947/boost-log-linking-error-on-osx#answer-38122244
   the order of libs is significant. I think the reordering of the libs
   should be safe for all platforms. Adding the "-mt" suffix was needed
   due to the way Homebrew installs the boost libs (there are no non-mt
   versions). This second one is probably trickier to get right in a
   portable manner, because it seems difficult to make assumptions we
   can hope will generally hold, on how people installed their boost
   libs.

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -318,12 +318,12 @@ if not conf.CheckLib ('boost_date_time', language = 'c++'):
 
 libs   = ['notmuch',
           'boost_filesystem',
-          'boost_system',
           'boost_program_options',
           'boost_log_setup',
-          'boost_log',
-          'boost_thread',
+          'boost_log-mt',
+          'boost_thread-mt',
           'boost_date_time',
+          'boost_system-mt',
           'stdc++']
 
 env.AppendUnique (LIBS = libs)
--- a/SConstruct
+++ b/SConstruct
@@ -320,10 +320,10 @@ libs   = ['notmuch',
           'boost_filesystem',
           'boost_program_options',
           'boost_log_setup',
-          'boost_log-mt',
-          'boost_thread-mt',
+          'boost_log',
+          'boost_thread',
           'boost_date_time',
-          'boost_system-mt',
+          'boost_system',
           'stdc++']
 
 env.AppendUnique (LIBS = libs)
--- a/src/utils/compiler.h
+++ b/src/utils/compiler.h
@@ -2,10 +2,28 @@
 
 # if defined (__clang__)
 
-# if __has_attribute(clang::fallthrough)
-# define FALLTHROUGH [[clang::fallthrough]]
+# ifndef __has_attribute
+#   define __has_attribute(x) 0
+# endif
+# ifndef __has_cpp_attribute
+#   define __has_cpp_attribute(x) 0
+# endif
+# ifndef __has_builtin
+#   define __has_builtin(x) 0
+# endif
+# ifndef __has_feature
+#   define __has_feature(x) 0
+# endif
+# ifndef __has_extension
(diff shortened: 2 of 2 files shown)


gauteh:
First one seems fine, but the second one I'm more reluctant to merge since I don't know if the `-mt` variant always works. Changing the order of the libs should be fine. Could you split this into three commits?
1. compiler.h fix
1. re-ordering fix
1. and possibly the -mt fix

the -mt fix might be better to do as a patch in the homebrew formula in that case.

Could you test if the meson build (#412) detects the correct boost libraries?

c-alpha:
Fully agree on the -mt fix. Should probably be fixed in the boost formula itself. It will be sufficient to symlink the -mt versions under non-mt names.

Since I have already pushed as a single commit, there are two options. If you really want commits only, I could delete my fork, create a new one and do three commits there before submitting a new PR for the new fork. The other option would be to back out the reordering and the -mt fixes from the current PR, and create two new branches. In each new branch, I would then the redordering and -mt fix respectively, and create two new PRs (one for each branch).

So three commits in a new PR, or two new PRs?

gauteh:
c-alpha writes on oktober 4, 2017 14:16:
> Fully agree on the -mt fix. Should probably be fixed in the boost formula itself. It will be sufficient to symlink the -mt versions under non-mt names.
> 
> Since I have already pushed as a single commit, there are two options. If you really want commits only, I could delete my fork, create a new one and do three commits there before submitting a new PR for the new fork. The other option would be to back out the reordering and the -mt fixes from the current PR, and create two new branches. In each new branch, I would then the redordering and -mt fix respectively, and create two new PRs (one for each branch).
> 
> So three commits in a new PR, or two new PRs?
> 

You can just force push to the current PR. If the only thing that should be 
merged is compiler.h changes then you don't need to re-send the other 
(unless you want to keep them for reference).

c-alpha:
The reordering is also needed, or else the link phase will fail. So I'll back out the -mt renaming, but leave the reordering in.

gauteh:
c-alpha writes on oktober 4, 2017 14:30:
> The reordering is also needed, or else the link phase will fail. So I'll back out the -mt renaming, but leave the reordering in.

Ah, right. Ok!

c-alpha:
Done, committed.

gauteh:
Thanks! Merged. I split this into two commits, before merging.

* dfd9e2a
* da25ea1

PR #428, opened 2017-11-04
Migrate to CMake build

c-alpha:
Use [CMake](https://cmake.org/) to build astroid. Currently, you can build astroid like this:
```
mkdir build
cmake -H. -Bbuild -DDISABLE_PLUGINS=ON
cmake --build build -- -j4
```

**Please submit any cmake-related issues to my branch at [c-alpha/astroid](https://github.com/c-alpha/astroid/issues).**

On the todo-list (no particular order):
- implement all `--disable_xxx` build options
- add a front-end shell script to save the rest of us from having to fiddle with the `cmake` command line
- many small things I can't think of right now, but which will need to be done


SConstruct
```diff
@@ -1,515 +0,0 @@
-import os, sys
```
gauteh: This should at least go in a separate commit so that we can easily decide when to delete SCons, but as mentioned - just leave both SCons and meson as is and we can decide when to remove them afterwards.

gauteh:
Nice work! Thans for this!!!

 Just to clarify; you don't need to `mkdir build` with that command line; you can also use the [ninja backend](https://ninja-build.org/) which was used with [meson](https://github.com/astroidmail/astroid/pull/412) and is [faster than the make backend](http://mesonbuild.com/Simple-comparison.html).

 So the most intuitive build steps are (using Ninja):
```sh
$ mkdir -p build
$ cd build
$ cmake -GNinja ..
$ ninja
```

while, the shortest version is:

```sh
$ cmake -H. -Bbuild -DDISABLE_PLUGINS=ON -GNinja
$ cmake --build build
```

> Ninja automatically runs the build in parallel.

gauteh:
@c-alpha: The remaining cmake features and issues should be solved in: https://github.com/c-alpha/astroid/pull/24. If anyone wants to test it please go ahead. Unless you see something specific I am going to merge this in not too long.

gauteh:
Merged into master! Thanks a lot @c-alpha for this huge effort !! Hopefully this should prove more stable than meson / scons. I plan to remove meson, but leave SCons untill cmake is proven to work.

c-alpha:
> Merged into master!

Whoahaha, we're going live!

> Thanks a lot @c-alpha for this huge effort !!

Wouldn't have been possible without your cooperation and contributions; "thanks!" goes back to you, too!

> Hopefully this should prove more stable than meson / scons.

I'm fairly confident that it will.

> I plan to remove meson, but leave SCons untill cmake is proven to work.

Sounds like a plan.

Issue #362, opened 2017-07-05
WYSIWYG HTML composer


bserrao:
It's possible to integrate an WYSIWYG HTML composer? Something like tinymce ou other program maybe.
I'm asking this because on my company everyone uses Outlook and writes HTML emails, and they don't seem to like that my replies break the "pretty" email chain.
Also is it possible to open the HTML part by default?

Thanks,

gauteh:
Related #236 and #319.

gauteh:
Don't have the possibility now, but if there is an editor that produces HTML then it could be used to generate the HTML part in the outgoing mail.

gauteh:
Would https://github.com/astroidmail/astroid/issues/319#issuecomment-470142756 solve the issue or does it need to be a fully featured HTML editor?

fakhir:
+1 on the fully featured HTML editor request. #319 doesn't cleanly handle replying to Outlook emails due to the intermediate html-to-markdown-to-html translation which doesn't work well for complex HTML emails such as those formatted by tools such as Jira.

PR #430, opened 2017-11-08
markdown: compose messages as markdown and send as multipart/alternative

gauteh:
html and text.

This requires a markdown to html converter like
[marked](https://github.com/chjj/marked). The markdown processor should
read from stdin and output html to stdout.

TODO:
  * Persistent markdown state over drafts (similar to signatures being
      added twic)
  * Always show HTML part when composing (would be great with a better
      UI for showing alternative parts #408)

> Remember that you can toggle the alternative part with Ctrl+Space when in EditMode (usually done with Enter).

__Requires [marked](https://github.com/chjj/marked) to be installed.__

Works as an alternative to: #362.

2017-11-08-125425_2560x1600_scrot

### DIFF
--- a/History.txt
+++ b/History.txt
@@ -1,5 +1,6 @@
 ==
 
+* Allow HTML parts to be composed using Markdown.
 * Use color to indicate whether a message is marked in thread-index.
 
 == v0.10.2 / 2017-10-11
--- a/meson.build
+++ b/meson.build
@@ -244,6 +244,7 @@ if not get_option('disable-tests')
 
   tests = [ [ 'generic test',   'test_generic',         [ 'tests/test_generic.cc' ]],
             [ 'compose testing', 'test_composed_message',     [ 'tests/test_composed_message.cc' ]],
+            [ 'markdown',       'test_markdown',        [ 'tests/test_markdown.cc' ]],
             [ 'non existant file', 'test_non_existant_file',  [ 'tests/test_non_existant_file.cc' ]],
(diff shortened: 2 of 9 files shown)


mxmehl:
Cool idea. Not sure whether I would use it personally though (I prefer plain tetx mails).

Do I understand the code correctly that this is optional and opt-in? Or should it become the default?

gauteh:
Totally optional of course, and off by default, though that could be configurable. Note that  the markdown source (which you input) is added as a plain text part, so astroid would show this by default.

yeled:
hooolllyyyy 😍

hugoroy:
Awesome! :-)

emdete:
very nice feature (while i am one of the pure text in emails guy as well)! 

it screws up the quoted stuff maybe some preprocessing would be helpful?

while my editor has markdown syntaxhighlighting it does not know about it (maybe the tempfile should have the extension ".md" to show that?)

gauteh:
M. Dietrich writes on februar 26, 2018 8:35:
> it screws up the quoted stuff maybe some preprocessing would be helpful?

Good point. Maybe the HTML with blockquote filter from GMime gives 
useful results?

> while my editor has markdown syntaxhighlighting it does not know about it (maybe the tempfile should have the extension ".md" to show that?)

Yes, should not create any big difficulties - the tmpfile is as far as I 
know only used during editing, but it is probably assumed that it is 
location is immutable.

PR #438, opened 2017-11-18
Message ID Generation Enhancements

c-alpha:
### Enhancements to the left of the `@`
- If `mail.message_id_user` configuration file parameter is empty, nothing is inserted. Previously, `.astroid` was forcibly inserted. To retain the same factory default behaviour, the default config file template has been updated to set this parameter to `astroid`. The `.` is still inserted if `mail.message_id_user` is not empty to retain compatibility with existing configurations.
- The generation of the random part has been updated to use the full set of allowed characters as per RFC 5322 to increase entropy and reduce risk of collisions. As a consequence, the message id can no longer be used as a file name for temp files. Hence, the generation of temp file names has been decoupled from the message id generation. Previously, external editors could recognize astroid temp files as email messages from the `@<fqdn>` naming convention (as the message id was used as the file name). To retain this scheme, a new config variable `editor.external_tmpfile_suffix` was introduced, with a default value of `@msg.draft`. Consequently, the version of the configuration file format had to be bumped up to 10. **This breaks compatibility with existing configurations** since the auto-detection in an external editor's configuration will likely need to be updated to recognize the new naming convention. Alternatively, if and only if the new `%1` parameter (cf. rhs changes below) is **not** used, the new `editor.external_tmpfile_suffix` parameter can be set to the same value as the `mail.message_id_fqdn` parameter in the astroid configuration.
### Enhancements to the right of the `@`
- The `mail.message_id_fqdn` can now contain a `%1`, which is dynamically replaced with the FQDN of the from address of the message being composed. The FQDN is updated whenever the from address is changed, either in an external editor or via the drop-down menu. This ensures that the FQDN part of the message id always matches the FQDN of the account used for sending.


src/modes/edit_message.cc
```diff
@@ -134,27 +135,41 @@ namespace Astroid {
...
+      //
+      // get value for right-hand side from config file
       ustring hostname = astroid->config ().get <string> ("mail.message_id_fqdn");
       UstringUtils::trim (hostname);
+      // if hostname contains a "%1", insert the FQDN of the default
+      // account's email address
+      if (hostname.find ("%1", 0) != std::string::npos) {
+        Account &a = (accounts->accounts)[accounts->default_account];
+        hostname = ustring::compose (hostname,
+                                     boost::replace_all_copy (string(fqdn_of_address(a.email)), "%", "%%"));
```
gauteh: The extra replace_all_copy is here to safeguard of messages where a.email might contain `%`. This might still be good to have.

src/modes/edit_message.cc
```diff
@@ -134,27 +135,41 @@ namespace Astroid {
...
-
+      //
+      // generate left-hand side of the message id
+      //
+      // get value for right-hand side from config file
       ustring hostname = astroid->config ().get <string> ("mail.message_id_fqdn");
       UstringUtils::trim (hostname);
+      // if hostname contains a "%1", insert the FQDN of the default
+      // account's email address
+      if (hostname.find ("%1", 0) != std::string::npos) {
```
gauteh: If there are no %1 the compose below will just silently not do any replacements, I believe you don't need to have this check then?

src/modes/edit_message.cc
```diff
@@ -719,15 +738,50 @@ namespace Astroid {
...
+    }
+  }
+
+  ustring EditMessage::fqdn_of_address (ustring _addr) {
+    size_t atpos = _addr.find("@");
+    ustring fqdn;
+
+    if ( ! _addr.empty()) {
+      fqdn = _addr.substr(atpos + 1, _addr.length() - atpos);
+    }
```
gauteh: What happens with invalid e-mail addresses not containing any @ ? Does this work on a local e-mail address? From root to charlie?

src/modes/edit_message.cc
```diff
@@ -719,15 +738,50 @@ namespace Astroid {
       account_no ++;
     }
 
+    update_message_id_rhs_from_sender(a->email);
```
gauteh: This might cause problems if this is work on an existing draft which exists in notmuch since the message is identified in notmuch by its ID. In that case the message would have to be move/copied in notmuch.

src/config.cc
```diff
@@ -194,6 +194,7 @@ namespace Astroid {
     default_config.put ("editor.cmd", "gvim -f -c 'set ft=mail' '+set fileencoding=utf-8' '+set enc=utf-8' '+set ff=unix' '+set fo+=w' %1");
     default_config.put ("editor.external_editor", true); // should be true on Wayland
 # endif
+    default_config.put ("editor.external_tmpfile_suffix", "@msg.draft"); // suffix of tmp files passed to external editors; default: '@msg.draft'
```
gauteh: Do we really need this? Is this for editors to automatically set filetype? E.g. the gvim line above sets ft=mail by force anyway. Does this really add any security; couldn't an app easily just use `lsof` to figure out that this is an email file anyway. The same with a draft email which is stored in a users maildir.

c-alpha: Two reasons:
1. It echoes the previous behaviour where the message id was used as the temp file name.
2. It helps `emacslient` (which I use ;-) to automatically set filetype

src/modes/edit_message.cc
c-alpha: You're right, in principle the check is not needed. By having the `if` whe can avoid having to access the default account  and having boost strings perform its little dance each and every time.

No strong feelings about this one.

src/config.cc
gauteh: Does the tmpfile change name if you change from address? What if you change the from address from within the editor with the file open? I though the point was to hide that this was a e-mail message, with this it will be just as easy to detect.

src/modes/edit_message.cc
gauteh: Ok, I don't see any immediate problem. The only issue is if there is going to be a `%2` in the future.

src/config.cc
gauteh: The suffix should probably be `.eml` in that case. Also; perhaps the best would be to completely decouple the file-name from message-id.

src/modes/edit_message.cc
c-alpha: This only happens for drafts which have not been sent. After a message has been sent, the message id is never touched again, of course.

For as long as I keep editing it within astroid (and maybe an external editor), the message will only exist as a tempfile in `~/.cache/astroid/`, and everything works fine. The sent message is filed in the sent maildir, but gets the message id as it's filename.

If I save the draft before sending, and change the message id, astroid crashes with a message of "[error] db: could not find message". So it seems notmuch has a built-in assumption that the filename is identical to the message id (and never checks the message id inside the file again).

So I **think** changing the message id **could** work if I manage to do three things:
1. use the real message id as the filename when drafting a message
2. update the file name when the message id is updated
3. have notmuch rescan the the files if and when the filename of the draft has been updated

I'll make a few experiments and keep you posted.

gauteh:
c-alpha writes on november 20, 2017 22:15:
> If I save the draft before sending, and change the message id, astroid crashes with a message of "[error] db: could not find message". So it seems notmuch has a built-in assumption that the filename is identical to the message id (and never checks the message id inside the file again).

No, we are adding the message to notmuch when saving as draft - the 
message is identified by message-id. The filename is less important, 
but if the message-id is changed as well it will appear as a new message 
to notmuch. It is possible to add tags to drafts, these would have to be 
copied over if we remove/add the draft on message-id change.

I think that if we are going to make the message-id mutable we should 
use a decoupled file-name, e.g: astroid-@msgtime@,@maildir-suffix@ for 
drafts stored in maildir. And something mktmp'y for working drafts.

src/modes/edit_message.cc
c-alpha: Ok, I got as far as decoupling all file name generation (including for storing in `MAILDIR`) from the message id. It seems to not cause any major mayhem, except one small quirk (but which shows up in the debug log only). See last commit 7b76ff66f5bdd076483eb1590c511ca9527f05a5.

The test run is as follows:
1. configure astroid to use an external editor, and start composing a new message
2. when back in astroid, save the new message as a draft
3. revisit the draft, and edit it once again in the external editor
4. in the external editor, change the `From:` address to another account of yours
5. when back in astroid, observe how the from address drop-down button has changed and shows the new account
6. send the message without further ado
7. dig into `MAILDIR/sent/cur` (or whtever you've configured it to) to discover the message id of the message you just sent
8. return to astroid and search for the message of the message just sent (`o`, `id:...`); there will be a search result

So notmuch seems to pick up the new message fine.

Where the cogwheels start making grinding noises in the debug log is in/around `on_message.cc`. Right after successful sending, `AddSentMessage::emit` is called with the **new** message id. This is followed by three db search operations. Then, `OnMessageAction::emit` is called, but with the **old** message id. Most of the times, nothing else bad happens, but occasionally it causes a crash somewhere in glibmm when searching for the thread relating to the old message id.

It would thus seem that somewhere in/around `on_message.cc` I would need to check whether the message id has been updated, and take note of it in the appropriate member variable. As mentioned above, notmuch already seems to grasp the new message id.

c-alpha:
> I think that if we are going to make the message-id mutable we should
> use a decoupled file-name, e.g: astroid-@msgtime@,@maildir-suffix@ for
> drafts stored in maildir. And something mktmp'y for working drafts.

Done (7b76ff66f5bdd076483eb1590c511ca9527f05a5).

The full spec of maildir is here: https://cr.yp.to/proto/maildir.html (but you probably already know that one by heart, don'y you? ;-)
It speaks about three filename parts joined together with dots (`.`):
- **left part:** time units since the epoch
- **middle part:** random-ish value to safeguard against two messages created by the same machine within the same time unit
- **right part:** hostname of the machine writing the file

All this is done to ensure unique file names, even when multiple machines spit files into the same directory across network mounts. This spec is in no way written as a requirement, and as you said, notmuch does not really care about the details of the filename construction other than it be unique across `MAILDIR`.

src/modes/edit_message.cc
c-alpha: I'll meditate over this one and think about suitable safeguards.

c-alpha:
>  I could do this for you if you check the box allow changes from maintainers when creating the PR.

This appears to have already been enabled. Did you try pushing?

gauteh:
c-alpha writes on november 23, 2017 19:33:
>>  I could do this for you if you check the box allow changes from maintainers when creating the PR.
> 
> This appears to have already been enabled. Did you try pushing?

Yes, I rebased and pushed. Did you have any non-merge related changes in the 
merge commit? They may (hopefully not) have been overwritten.. (make a 
backup of your local branch first in that case!).

- gaute

src/compose_message.cc
```diff
@@ -608,7 +608,8 @@ namespace Astroid {
 
         if (account->save_sent) {
           using bfs::path;
-          save_to = account->save_sent_to / path(id + ":2,");
+          // canonical maildir spec: https://cr.yp.to/proto/maildir.html
+          save_to = account->save_sent_to / path(UstringUtils::random_file_name (":2,"));
```
gauteh: Does this add the timestamp first ?

c-alpha:
On 2017-11-24, at 10:32 , Gaute Hope <notifications@github.com> wrote:

> c-alpha writes on november 23, 2017 19:33:
> >> I could do this for you if you check the box allow changes from maintainers when creating the PR.
> > 
> > This appears to have already been enabled. Did you try pushing?
> 
> Yes, I rebased and pushed. Did you have any non-merge related changes in the 
> merge commit?

Umm, no.

> They may (hopefully not) have been overwritten.. (make a 
> backup of your local branch first in that case!).
> [...]

It worked. Thanks!

src/modes/edit_message.cc
gauteh: Notmuch doesn't pick up anything automatically. It either has to be be set in a consistent state by _database_remove/_database_index calls or `notmuch new` must be run. We should not rely on the latter, and the database should always be in a consistent state after astroid has done any modifications. The `OnMessage` error should probably not be fixed in `OnMessage`, it likely stems from some out-of-date message-id somewhere else and should be fixed there (since previously it was safe to assume that the message-id was immutable).

src/modes/edit_message.cc
```diff
@@ -719,15 +738,50 @@ namespace Astroid {
...
+    
     bool same_account = (rn == from_combo->get_active_row_number ());
     LOG (debug) << "same account: " << same_account;
     if (!same_account) {
       reset_signature ();
     }
-
+
```
gauteh: ws.

src/utils/ustring_utils.cc
```diff
@@ -53,19 +52,39 @@ namespace Astroid {
...
+    uint32_t r;
 
-    const string _chars = "abcdefghijklmnopqrstuvwxyz1234567890";
-    random_device rd;
-    mt19937 g(rd());
-
-    for (int i = 0; i < length; i++)
-      str += _chars[g() % _chars.size()];
+    for (int i = 0; i < length; i++) {
+      r = arc4random_uniform(chars.length());   // This is recommended over constructions
```
gauteh: Seems to be missing an include here?

src/utils/ustring_utils.cc
```diff
@@ -53,19 +52,39 @@ namespace Astroid {
...
   }
+  
+  ustring UstringUtils::random_alphanumeric (int length) {
+    const ustring _chars = "abcdefghijklmnopqrstuvwxyz1234567890";
+    return UstringUtils::random_string_from_charset(_chars, length);
+  }
 
+  ustring UstringUtils::random_rfc5322_atext (int length) {
+    // 81 allowed charaters for the message id as per RFC 5322
+    const ustring _chars = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz1234567890!#$%&'*+-/=?^_`{|}~";
```
gauteh: Maybe it would be safer(nicer to skip some of those special ones which could easily cause shell-trouble and file-name issues (if used in that way).

gauteh:
c-alpha writes on november 24, 2017 10:42:
> On 2017-11-24, at 10:32 , Gaute Hope <notifications@github.com> wrote:
> 
>> c-alpha writes on november 23, 2017 19:33:
>> >> I could do this for you if you check the box allow changes from maintainers when creating the PR.
>> > 
>> > This appears to have already been enabled. Did you try pushing?
>> 
>> Yes, I rebased and pushed. Did you have any non-merge related changes in the 
>> merge commit?
> 
> Umm, no.
> 
>> They may (hopefully not) have been overwritten.. (make a 
>> backup of your local branch first in that case!).
>> [...]
> 
> It worked. Thanks!
> 

Ok good! I got confused about the commit that you mentioned about the 
mktmp-stuff, maybe the SHA changed at a later point..

I realize that some of the comments I made will make it possible for 
someone to detect the MUA with a large enough sample of emails.. I think 
this is really hard to safe-guard completely against anyway. There are 
probably other quirks that are identifiabel as well.

src/utils/ustring_utils.cc
c-alpha: Hm, compiles for me. Do you get an error? Usually it's in `stdlib.h`.

src/utils/ustring_utils.cc
c-alpha: No longer used as filename.

In the shell, you'll have to enclose it in double quotes.

gauteh:
c-alpha writes on november 24, 2017 11:00:
> c-alpha commented on this pull request.
> 
> 
> 
>>  
> -    const string _chars = "abcdefghijklmnopqrstuvwxyz1234567890";
> -    random_device rd;
> -    mt19937 g(rd());
> -
> -    for (int i = 0; i < length; i++)
> -      str += _chars[g() % _chars.size()];
> +    for (int i = 0; i < length; i++) {
> +      r = arc4random_uniform(chars.length());   // This is recommended over constructions
> 
> Hm, compiles for me. Do you get an error? Usually it's in `stdlib.h`.

No - compiles fine I think, the header is probably included somewhere, 
just not in ustring_utils.cc yet. Good practice to include it here as 
well.

gauteh:
c-alpha writes on november 24, 2017 11:01:
> c-alpha commented on this pull request.
> 
> 
> 
>>  
> +  ustring UstringUtils::random_rfc5322_atext (int length) {
> +    // 81 allowed charaters for the message id as per RFC 5322
> +    const ustring _chars = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz1234567890!#$%&'*+-/=?^_`{|}~";
> 
> No longer used as filename.
> 
> In the shell, you'll have to enclose it in double quotes.

Yes, in the notmuch world the message-id is used a lot, also from the 
command line. I think we should stick to chars that don't require 
quoting.

c-alpha:
> I realize that some of the comments I made will make it possible for
> someone to detect the MUA with a large enough sample of emails.. I think
> this is really hard to safe-guard completely against anyway.

Not hard to figure out the MUA as long as the `User-Agent:` or `X-Mailer:` headers are included... ;-)

> There are probably other quirks that are identifiabel as well.

Identifying the MUA can either hint you to exploitable vulnerabilities of it, or in case it is a not very widely used one, can be used in social engineering.

Social engineering is also the motivation for my attempt at having a different message id scheme for each account, btw. If there is a fixed part, it helps figuring out under which names I am sending email. If the message id rhs is identical to the `From:` domain name, it doesn't give any additional hints for social engineering.

There are a great many schemes in use for generating message id. A few real world examples:
- Microsoft Outlook 16.0 `Message-ID: <07b301d26c10$027b5780$07720680$@teletrust.de>`
- Smart_Send_4_1_3 `Message-ID: <5716441053928186124524@PC81265>`
- microtech's ERP-complete Version ERP-17.0.5626 BE `Message-ID: <9668898b-32f7-4eef-8fe1-427a23ddb488@VMSERVER5.jacob.local>`
- Zendesk Mailer `Message-ID: <W5MKP04XK5_599d8e47c3409_12c663ffcc56cb98c2026d6_sprut@zendesk.com>`
- PHPMailer 5.2.17 `Message-ID: <a01c10f8267c198b6fb44d398fdcd53a@www.digitalo.de>`
- Apple Mail `Message-Id: <4F6D4258-0BDB-4C62-AD23-495D80FA524D@gmx.de>`
- DocBridgePilot `Message-id: <1707212701.151015.1482230319892.JavaMail.docadm.prd@SVSAPDOCP01>`
- David.fx by Tobit.Software `Message-ID: <00062124.57D98266@172.20.19.216>`
- unknown `Message-ID: <1715526924.248001.1510267708213.JavaMail.reomf@qde7xg>`
- unknown `Message-Id: <fab2dab0.ADkAADwHFGIAAUhWEYMAAF70hJsAANqOvFgAAAAAAAFZIgBZ-dc4@mailjet.com>`
- unknown `Message-ID: <SEMA-CR-1-4VXXGSD@bmsmail2.ieee.org>`
- unknown `Message-Id: <20171016152113.514FA1828247@svr5-vm.m-online.net>`
- unknown `Message-ID: <bRiHy000000000000000000000000000000000000000000000OUIVM000o3IO7GnaSTi8epIBoz5hoA@sfdc.net>`
- unknown `Message-ID: <61BFC5496BF7D24CAC742F4D6F532CC0162C101A@S-EU-MUC01MXB12.eu.accor.net>`

Observations:
- many people are disclosing more information than they probably would want to
- JavaMail is the only one to seemingly reuse the maildir file name in the message id (and I don't think that would be a smart idea as it gives you a local file name for free)
- MS Outlook seems to do the same if you replace the `$` by `.`
- UUID is seeing some popularity

Discussion:
With my currently proposed scheme, astroid would clearly stick out because of the punctuation characters being used. The length of the lhs can also be indicative, so might also need to be randomized. If we want to buy into the "let's fool the attacker" fray, as astroid will all but certainly be used on unixoid systems only, we could mimic the scheme used by desktop Outlook (cf. first example)? Or I could implement several mimicking schemes, and pick one at random for each message?

src/modes/edit_message.cc
c-alpha: > Notmuch doesn't pick up anything automatically.

I know. So it seems that the existing code updates the db often enough already?

gauteh:
c-alpha writes on november 24, 2017 12:36:
> c-alpha commented on this pull request.
> 
> 
> 
>> @@ -719,15 +738,50 @@ namespace Astroid {
>        account_no ++;
>      }
>  
> +    update_message_id_rhs_from_sender(a->email);
> 
>> Notmuch doesn't pick up anything automatically.
> 
> I know. So it seems that the existing code updates the db often enough already?
> 

Ok, please also test with:

* Create new message
* Save as draft
* Add a tag to draft
* Re-open message
* Make changes including changing the from address (either from within 
  editor or not)
* See if draft message still exists and has its tags or anything else 
  broke.

gauteh:
c-alpha writes on november 24, 2017 12:19:
>> I realize that some of the comments I made will make it possible for
>> someone to detect the MUA with a large enough sample of emails.. I think
>> this is really hard to safe-guard completely against anyway.
> 
> Not hard to figure out the MUA as long as the `User-Agent:` or `X-Mailer:` headers are included... ;-)

Ok, I don't think we should bother with trying to spoofing the attacker 
(that could go into a plugin if someone is interested). I thought that 
making the UA hidden was part of your motivation, but now I understand 
it to be:

* Not being able to readily link messages sent using different accounts to the same author.
* Have different fqdn's for different accounts for other reasons as well
* Not leaking local hostname.

c-alpha:
On 2017-11-24, at 12:38 , Gaute Hope <notifications@github.com> wrote:

> [...]
> Ok, please also test with:
> 
> * Create new message
> * Save as draft
> * Add a tag to draft
> * Re-open message
> * Make changes including changing the from address (either from within 
> editor or not)
> * See if draft message still exists and has its tags or anything else 
> broke.
> [...]

Tag is retained, all seems fine. Nothing suspicious in the debug log either.

c-alpha:
On 2017-11-24, at 12:40 , Gaute Hope <notifications@github.com> wrote:

> [...]
> Ok, I don't think we should bother with trying to spoofing the attacker 
> (that could go into a plugin if someone is interested).

Good point!

> I thought that 
> making the UA hidden was part of your motivation, but now I understand 
> it to be:
> 
> * Not being able to readily link messages sent using different accounts to the same author.
> * Have different fqdn's for different accounts for other reasons as well
> * Not leaking local hostname.
> [...]

To the point.

src/modes/edit_message.cc
c-alpha: > The OnMessage error should probably not be fixed in OnMessage, it likely stems from some out-of-date message-id somewhere else and should be fixed there (since previously it was safe to assume that the message-id was immutable).

I am currently trying to figure out where this somewhere else could be.

I have added a `LOG` statement in `void Message::on_message_updated (Db * db, ustring _mid)` which maybe can shed a little more light. The log excerpt is below. The first "emitted updated signal for message:" is followed by three invocations of `Message::on_message_updated`, where the last one features the new message id. The second "emitted updated signal for message:" is followed by two invocations of `Message::on_message_updated`, but before the third invocation occurs, there is a "db: could not find message:", followed by a single Message::on_message_updated` with the old message id. Then it goes on to crash.

```
[15:27:30.060336] [0x00007000093a8000] [debug] db: open time: 94.149 ms.
[15:27:30.060437] [0x00007000093a8000] [info] db: adding sent message: /Users/alexanderadolf/Library/Mail/Maildir/sent/cur/5si6iga31o0m6fdo9ablfyd92akuhus8yzilnk5fj7ogu6pjee:2,
[15:27:30.071548] [0x00007000093a8000] [info] db: closing db.
[15:27:30.091940] [0x00007000093a8000] [debug] db: rw: releasing lock.
[15:27:30.091991] [0x00007000093a8000] [debug] db: rw-s: releasing lock.
[15:27:30.092027] [0x00007000093a8000] [info] db: open db read-write.
[15:27:30.092047] [0x00007000093a8000] [debug] db: rw-s: waiting for rw lock.. (r-o open: 0)
[15:27:30.092100] [0x00007000093a8000] [debug] db: rw-s lock acquired.
[15:27:30.191504] [0x00007000093a8000] [debug] db: open time: 110.952 ms.
[15:27:30.195270] [0x00007000093a8000] [info] db: closing db.
[15:27:30.215011] [0x00007000093a8000] [debug] db: rw: releasing lock.
[15:27:30.215063] [0x00007000093a8000] [debug] db: rw-s: releasing lock.
[15:27:30.262033] [0x00007fffad7a0340] [info] db: open db read-only, waiting for lock..
[15:27:30.262096] [0x00007fffad7a0340] [debug] db: read-only got lock.
[15:27:30.262355] [0x00007fffad7a0340] [debug] db: open time: 0.331 ms.
[15:27:30.262401] [0x00007fffad7a0340] [info] actions: emitted updated signal for message: I%9Xu/8c8wy2'Xz%|~MKA8urvpiK}187kV/+fHW^@mnet-online.de
[15:27:30.262441] [0x00007fffad7a0340] [debug] msg: on_message_updated: mid = I%9Xu/8c8wy2'Xz%|~MKA8urvpiK}187kV/+fHW^@condition-alpha.com / _mid = I%9Xu/8c8wy2'Xz%|~MKA8urvpiK}187kV/+fHW^@mnet-online.de
[15:27:30.262474] [0x00007fffad7a0340] [debug] msg: on_message_updated: mid = I%9Xu/8c8wy2'Xz%|~MKA8urvpiK}187kV/+fHW^@condition-alpha.com / _mid = I%9Xu/8c8wy2'Xz%|~MKA8urvpiK}187kV/+fHW^@mnet-online.de
[15:27:30.262507] [0x00007fffad7a0340] [debug] msg: on_message_updated: mid = I%9Xu/8c8wy2'Xz%|~MKA8urvpiK}187kV/+fHW^@mnet-online.de / _mid = I%9Xu/8c8wy2'Xz%|~MKA8urvpiK}187kV/+fHW^@mnet-online.de
[15:27:30.262592] [0x00007fffad7a0340] [info] actions: emitted changed signal for thread: 0000000000000296
[15:27:30.262621] [0x00007fffad7a0340] [debug] searches: refreshing..
[15:27:30.262643] [0x00007fffad7a0340] [debug] searches: skipping, not visible.
[15:27:30.262659] [0x00007fffad7a0340] [info] ql (0): tag:inbox, got changed thread signal: 0000000000000296
[15:27:30.262712] [0x00007fffad7a0340] [debug] db: checking if thread: 0000000000000296 matches query: tag:inbox
[15:27:30.262825] [0x00007fffad7a0340] [debug] db: thread in query check: 0.12 ms.
[15:27:30.262850] [0x00007fffad7a0340] [debug] ql: updated: did not find thread, time used: 0.177 ms.
[15:27:30.262867] [0x00007fffad7a0340] [info] ql (1): tag:draft, got changed thread signal: 0000000000000296
[15:27:30.262924] [0x00007fffad7a0340] [debug] db: checking if thread: 0000000000000296 matches query: tag:draft
[15:27:30.268751] [0x00007fffad7a0340] [debug] db: thread in query check: 0.689 ms.
[15:27:30.268903] [0x00007fffad7a0340] [debug] ql: updated: did not find thread, time used: 0.864 ms.
[15:27:30.268971] [0x00007fffad7a0340] [info] ql (2): tag:draft, got changed thread signal: 0000000000000296
[15:27:30.269106] [0x00007fffad7a0340] [debug] db: checking if thread: 0000000000000296 matches query: tag:draft
[15:27:30.269405] [0x00007fffad7a0340] [debug] db: thread in query check: 0.316 ms.
[15:27:30.269468] [0x00007fffad7a0340] [debug] ql: updated: did not find thread, time used: 0.454 ms.
[15:27:30.269539] [0x00007fffad7a0340] [info] db: closing db.
[15:27:30.269733] [0x00007fffad7a0340] [debug] db: ro: waiting for lock to close..
[15:27:30.269789] [0x00007fffad7a0340] [debug] db: ro: closing..
[15:27:30.269869] [0x00007fffad7a0340] [info] db: open db read-only, waiting for lock..
[15:27:30.269912] [0x00007fffad7a0340] [debug] db: read-only got lock.
[15:27:30.270327] [0x00007fffad7a0340] [debug] db: open time: 0.441 ms.
[15:27:30.270410] [0x00007fffad7a0340] [info] actions: emitted updated signal for message: I%9Xu/8c8wy2'Xz%|~MKA8urvpiK}187kV/+fHW^@condition-alpha.com
[15:27:30.270481] [0x00007fffad7a0340] [debug] msg: on_message_updated: mid = I%9Xu/8c8wy2'Xz%|~MKA8urvpiK}187kV/+fHW^@condition-alpha.com / _mid = I%9Xu/8c8wy2'Xz%|~MKA8urvpiK}187kV/+fHW^@condition-alpha.com
[15:27:30.270543] [0x00007fffad7a0340] [debug] msg: on_message_updated: mid = I%9Xu/8c8wy2'Xz%|~MKA8urvpiK}187kV/+fHW^@condition-alpha.com / _mid = I%9Xu/8c8wy2'Xz%|~MKA8urvpiK}187kV/+fHW^@condition-alpha.com
[15:27:30.270751] [0x00007fffad7a0340] [error] db: could not find message: I%9Xu/8c8wy2'Xz%|~MKA8urvpiK}187kV/+fHW^@condition-alpha.com, status: 0
[15:27:30.270827] [0x00007fffad7a0340] [info] message: emitted changed signal for message: I%9Xu/8c8wy2'Xz%|~MKA8urvpiK}187kV/+fHW^@condition-alpha.com: 0
[15:27:30.270884] [0x00007fffad7a0340] [debug] msg: on_message_updated: mid = I%9Xu/8c8wy2'Xz%|~MKA8urvpiK}187kV/+fHW^@mnet-online.de / _mid = I%9Xu/8c8wy2'Xz%|~MKA8urvpiK}187kV/+fHW^@condition-alpha.com
[15:27:30.270961] [0x00007fffad7a0340] [error] db: could not find message: I%9Xu/8c8wy2'Xz%|~MKA8urvpiK}187kV/+fHW^@condition-alpha.com, status: 0
[15:27:30.271047] [0x00007fffad7a0340] [info] actions: emitted updated and changed signal for thread: 0000000000000295
[15:27:30.271341] [0x00007fffad7a0340] [error] db: could not find thread: 0000000000000295
[15:27:30.271480] [0x00007fffad7a0340] [info] db: closing db.
[15:27:30.271575] [0x00007fffad7a0340] [debug] db: ro: waiting for lock to close..
[15:27:30.271612] [0x00007fffad7a0340] [debug] db: ro: closing..

(<unknown>:37441): glibmm-ERROR **:
what: db: could not find thread!

Trace/BPT trap: 5
```

When running it in the debugger, the stack backtrace is this:
```
(lldb) bt
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BREAKPOINT (code=EXC_I386_BPT, subcode=0x0)
  * frame #0: 0x00000001031e2f17 libglib-2.0.0.dylib`_g_log_abort + 22
    frame #1: 0x00000001031e2da2 libglib-2.0.0.dylib`g_log_default_handler + 206
    frame #2: 0x00000001031e2a5e libglib-2.0.0.dylib`g_logv + 878
    frame #3: 0x00000001031e26d2 libglib-2.0.0.dylib`g_log + 120
    frame #4: 0x000000010393c6cd libglibmm-2.4.1.dylib`Glib::exception_handlers_invoke() + 303
    frame #5: 0x000000010393b064 libglibmm-2.4.1.dylib`Glib::DispatchNotifier::pipe_io_handler(Glib::IOCondition) + 326
    frame #6: 0x000000010393f13c libglibmm-2.4.1.dylib`Glib::IOSource::dispatch(sigc::slot_base*) + 52
    frame #7: 0x000000010393e253 libglibmm-2.4.1.dylib`Glib::Source::dispatch_vfunc(_GSource*, int (*)(void*), void*) + 39
    frame #8: 0x00000001031dce64 libglib-2.0.0.dylib`g_main_context_dispatch + 258
    frame #9: 0x00000001031dd162 libglib-2.0.0.dylib`g_main_context_iterate + 426
    frame #10: 0x00000001031dd1ba libglib-2.0.0.dylib`g_main_context_iteration + 55
    frame #11: 0x0000000102bafc56 libgio-2.0.0.dylib`g_application_run + 468
    frame #12: 0x000000010001e839 astroid`Astroid::Astroid::run(this=0x0000000106d00480, argc=1, argv=0x00007ffeefbff530) at astroid.cc:320
    frame #13: 0x000000010000d5aa astroid`main(argc=1, argv=0x00007ffeefbff530) at main.cc:5
    frame #14: 0x00007fff74150145 libdyld.dylib`start + 1
```

Does any of the above ring any bells with you where I should be taking a closer look?

c-alpha:
I finally got to take care of this again.

Now, I'm getting Boost version warnings when running `cmake`:
```
CMake Warning at /usr/local/Cellar/cmake/3.10.2/share/cmake/Modules/FindBoost.cmake:801 (message):
  New Boost version may have incorrect or missing dependencies and imported
  targets
Call Stack (most recent call first):
  /usr/local/Cellar/cmake/3.10.2/share/cmake/Modules/FindBoost.cmake:907 (_Boost_COMPONENT_DEPENDENCIES)
  /usr/local/Cellar/cmake/3.10.2/share/cmake/Modules/FindBoost.cmake:1542 (_Boost_MISSING_DEPENDENCIES)
  CMakeLists.txt:102 (find_package)
```
And when I run the executable I built, it crashes on a missing symbol:
```
dyld: Symbol not found: _UCNV_FROM_U_CALLBACK_ESCAPE_59
  Referenced from: /usr/local/opt/webkitgtk@2.4.11/lib/libwebkitgtk-3.0.0.dylib
  Expected in: flat namespace
 in /usr/local/opt/webkitgtk@2.4.11/lib/libwebkitgtk-3.0.0.dylib
Abort trap: 6
```
Are those two issues related? Any hints appreciated. Please don't tell me I need to rebuild webkitgtk@2.4.11... @gauteh ?

gauteh:
tor. 1. feb. 2018 kl. 17:48 skrev c-alpha <notifications@github.com>:

> I finally got to take care of this again.
>
> Now, I'm getting Boost version warnings when running cmake:
>
> CMake Warning at /usr/local/Cellar/cmake/3.10.2/share/cmake/Modules/FindBoost.cmake:801 (message):
>   New Boost version may have incorrect or missing dependencies and imported
>   targets
> Call Stack (most recent call first):
>   /usr/local/Cellar/cmake/3.10.2/share/cmake/Modules/FindBoost.cmake:907 (_Boost_COMPONENT_DEPENDENCIES)
>   /usr/local/Cellar/cmake/3.10.2/share/cmake/Modules/FindBoost.cmake:1542 (_Boost_MISSING_DEPENDENCIES)
>   CMakeLists.txt:102 (find_package)
>
> And when I run the executable I built, it crashes on a missing symbol:
>
> dyld: Symbol not found: _UCNV_FROM_U_CALLBACK_ESCAPE_59
>   Referenced from: /usr/local/opt/webkitgtk@2.4.11/lib/libwebkitgtk-3.0.0.dylib
>   Expected in: flat namespace
>  in /usr/local/opt/webkitgtk@2.4.11/lib/libwebkitgtk-3.0.0.dylib
> Abort trap: 6
>
> Are those two issues related? Any hints appreciated. Please don't tell me
> I need to rebuild webkitgtk@2.4.11... @gauteh <https://github.com/gauteh>
> ?
>
>
Does not look too good.. maybe somewhat related to #459?

Maybe you have to start with a fresh build dir. I am making fast progress
with webkit2 extension version. So we need to have WebKit2 on Mac soon too.

c-alpha:
> Does not look too good.. maybe somewhat related to #459?

Seems quite likely.

> Maybe you have to start with a fresh build dir. I am making fast progress
> with webkit2 extension version. So we need to have WebKit2 on Mac soon too.

Fresh build is no prob. Which branch should I check out?

gauteh:
tor. 1. feb. 2018 kl. 18:34 skrev c-alpha <notifications@github.com>:

> Does not look too good.. maybe somewhat related to #459
> <https://github.com/astroidmail/astroid/issues/459>?
>
> Seems quite likely.
>
> Maybe you have to start with a fresh build dir. I am making fast progress
> with webkit2 extension version. So we need to have WebKit2 on Mac soon too.
>
> Fresh build is no prob. Which branch should I check out?
>

Think it is the currently top PR.

c-alpha:
> Think it is the currently top PR.

#455 Webkit2ext ?

gauteh:
c-alpha writes on februar 2, 2018 15:14:
>> Think it is the currently top PR.
> 
> #455 Webkit2ext ?

Yes

c-alpha:
I cloned `gauteh-astroid/webkit2ext`:

```
-- Checking for module 'webkit2gtk-4.0'
--   No package 'webkit2gtk-4.0' found
CMake Error at /usr/local/Cellar/cmake/3.10.2/share/cmake/Modules/FindPkgConfig.cmake:415 (message):
  A required package was not found
Call Stack (most recent call first):
  /usr/local/Cellar/cmake/3.10.2/share/cmake/Modules/FindPkgConfig.cmake:593 (_pkg_check_modules_internal)
  CMakeLists.txt:100 (pkg_check_modules)
```
Ha! I presume there is no homebrew tap for `webkit2gtk-4.0` yet?

gauteh:
c-alpha writes on februar 2, 2018 17:54:
> I cloned `gauteh-astroid/webkit2ext`:
> 
> ```
> -- Checking for module 'webkit2gtk-4.0'
> --   No package 'webkit2gtk-4.0' found
> CMake Error at /usr/local/Cellar/cmake/3.10.2/share/cmake/Modules/FindPkgConfig.cmake:415 (message):
>   A required package was not found
> Call Stack (most recent call first):
>   /usr/local/Cellar/cmake/3.10.2/share/cmake/Modules/FindPkgConfig.cmake:593 (_pkg_check_modules_internal)
>   CMakeLists.txt:100 (pkg_check_modules)
> ```
> Ha! I presume there is no homebrew tap for `webkit2gtk-4.0` yet?

That's the crux! https://github.com/astroidmail/homebrew-astroid/issues/11

I think @yeled hoped macports (or whatever it's called) would fix it.

c-alpha:
> That's the crux! astroidmail/homebrew-astroid#11

> I think @yeled hoped macports (or whatever it's called) would fix it.

Macports is a competitor of Homebrew. I myself was a long time user of Macports, but have switched to Homebrew. For reasons. So please choose one or the other, but please don't require both.

In the simplest case, only a new formula would need to be added to [astroidmail/homebrew-astroid](https://github.com/astroidmail/homebrew-astroid/tree/master/Formula), next  to the existing one for the old `webkitgtk@2.4.11.rb` version, and then `astroid.rb` updated accordingly.

gauteh:
c-alpha writes on februar 5, 2018 15:09:
>> That's the crux! astroidmail/homebrew-astroid#11
> 
>> I think @yeled hoped macports (or whatever it's called) would fix it.
> 
> Macports is a competitor of Homebrew. I myself was a long time user of Macports, but have switched to Homebrew. For reasons. So please choose one or the other, but please don't require both.
> 
> In the simplest case, only a new formula would need to be added to [astroidmail/homebrew-astroid](https://github.com/astroidmail/homebrew-astroid/tree/master/Formula), next  to the existing one for the old `webkitgtk@2.4.11.rb` version, and then `astroid.rb` updated accordingly.

I currently don't use either, but the hope was that macports would get 
webkit2 compiling on mac so that it would be easy to port to homebrew.

c-alpha:
> [...] but the hope was that macports would get webkit2 compiling on mac so that it would be easy to port to homebrew.

"Source of inspiration" type use. I see... ;-))

c-alpha:
I have changed my mind wrt. the `%1` [DWIM](https://en.wikipedia.org/wiki/DWIM)-magic for generating the right-hand side of the message id.

I am now reworking the code to make `message_id_user` and `message_id_fqdn` simple (read: no auto-replacement) parameters per account. That way, users can have the same values for two accounts, but different ones for all others, or pure phantasy values everywhere. Clearer end-user control, clearer code.

> “When I’m working on a problem, I never think about beauty. I think only how to solve the problem. But when I have finished, if the solution is not beautiful, I know it is wrong.”
> —R. Buckminster Fuller (1895–1983),
> American Architect, Author, Designer, Inventor, and Futurist

Just can't test at the moment due to the webkitgtk issue.

gauteh:
c-alpha writes on februar 5, 2018 17:04:
> I have changed my mind wrt. the `%1` [DWIM](https://en.wikipedia.org/wiki/DWIM)-magic for generating the right-hand side of the message id.
> 
> I am now reworking the code to make `message_id_user` and `message_id_fqdn` simple (read: no auto-replacement) parameters per account. That way, users can have the same values for two accounts, but different ones for all others, or pure phantasy values everywhere. Clearer end-user control, clearer code.
> 
>> “When I’m working on a problem, I never think about beauty. I think only how to solve the problem. But when I have finished, if the solution is not beautiful, I know it is wrong.”
>> —R. Buckminster Fuller (1895–1983),
>> American Architect, Author, Designer, Inventor, and Futurist
> 
> Just can't test at the moment due to the webkitgtk issue.
> 

That does sound good! Simple, but sufficiently complicated to solve the 
problem, is always better!

c-alpha:
@gauteh, it seems I need a hint or two from you. ;-)

All seems fine for as long as I fiddle with a draft message, and keep changing its `From:` address. I do however get a crash when I actually send such a message. It gets delivered fine, so the crash occurs right after sending. 

If and when I **do not** change the `From:` address at all, I see a log message like:
```
[info] actions: emitted updated signal for message: <message-id>
```
If and when I **do** change the `From:` address before sending, I see the same log message, but with a message id corresponding to the **old** `From:` address, and followed by:
```
[error] db: could not find message: <old-message-id>, status: 0
```
I peeked around the code a bit, and ended up staring at `onmessage.cc`. Could it be the case that `AddDraftMessage::doit()` and `AddSentMessage::doit()` (i.e. the only methods where the message id member variable gets set) are only ever called once (when the message is first created?), and when the corresponding `::emit()` methods are called, the message has received a new message id, but the `mid` member variable still has its initial value?

gauteh:
c-alpha writes on februar 14, 2018 22:19:
> @gauteh, it seems I need a hint or two from you. ;-)
> 
> All seems fine for as long as I fiddle with a draft message, and keep changing its `From:` address. I do however get a crash when I actually send such a message. It gets delivered fine, so the crash occurs right after sending. 
> 
> If and when I **do not** change the `From:` address at all, I see a log message like:
> ```
> [info] actions: emitted updated signal for message: <message-id>
> ```
> If and when I **do** change the `From:` address before sending, I see the same log message, but with a message id corresponding to the **old** `From:` address, and followed by:
> ```
> [error] db: could not find message: <old-message-id>, status: 0
> ```
> I peeked around the code a bit, and ended up staring at `onmessage.cc`. Could it be the case that `AddDraftMessage::doit()` and `AddSentMessage::doit()` (i.e. the only methods where the message id member variable gets set) are only ever called once (when the message is first created?), and when the corresponding `::emit()` methods are called, the message has received a new message id, but the `mid` member variable still has its initial value?
> 

Not sure; because drafts are added to the DB when saving them the first 
time. So if you are changing the mid this would actually create a new 
message + make the old message invalid (in the notmuch db). In fact; 
when re-saving an existing draft it should be migrated from the old mid 
to the new (in case a user has added tags to the draft message already). 
Does this happen with regular sent messages or only with drafts which 
have had their from address changed from the original one?

This has to be done now since message-ids of drafts have become mutable.

The mid comes from adding the file to the db, so that should be correct. 
Maybe the signal is emitted somewhere else on the old ID. Should 
probably create a new Action called SaveDraftMessage which migrates the 
old mid to the new.

c-alpha:
Thanks for your thuoghts, @gauteh!

> Does this happen with regular sent messages

No. 

> or only with drafts which have had their from address changed from the original one?

Exactly.

After some more `LOG (trace) << ...` debugging, I got to the following conclusion:

After the message has been sent, there is a series of calls to `AddSentMessage` methods, which use the new message id. **But** these are interleaved with two calls to `OnMessageAction` methods, which use the old message id, and which presumably lead to the crash.

I think that the  calls to `OnMessageAction` methods with the old message id originate from `EditMessage` trying to delete the draft message after it has been sent. `EditMessage` uses a `draft_msg` instance variable, for which it calls `delete_draft()` methods. It seems that `draft_msg` is only updated when the draft is first loaded from the file system, but not again when the message is edited before sending.

So I think I shall have a closer look at the handling of `draft_msg` in `EditMessage`.

gauteh:
c-alpha writes on februar 15, 2018 15:39:
> I think that the  calls to `OnMessageAction` methods with the old message id originate from `EditMessage` trying to delete the draft message after it has been sent. `EditMessage` uses a `draft_msg` instance variable, for which it calls `delete_draft()` methods. It seems that `draft_msg` is only updated when the draft is first loaded from the file system, but not again when the message is edited before sending.

That seems plausible. I don't think the mid has to be migrated in the db 
unless the message is saved - otherwise, as long as the draft hasn't 
been written back, the mid matches the file on disk (in db). Once the 
draft is saved the notmuch-message must be migrated (tags copied).

src/modes/edit_message.cc
```diff
@@ -666,7 +668,56 @@ namespace Astroid {
...
+//            draft_msg->message->message_id = (char*) msg_id.c_str();
+            ThreadView *tv = thread_view;
+            astroid->actions->doit(refptr<Action>(new OnMessageAction (draft_msg->mid, "" /*draft_msg->tid*/,
+              [fname, cmsg, tv] (Db * db, notmuch_message_t * msg) {
+                // first, remove any draft tags
+                for (ustring t : Db::draft_tags) {
+                  notmuch_message_remove_tag (msg, t.c_str ());
+                }
+                // remember remaining (non-draft) tags (if any)
+                notmuch_tags_t *keep_tags = notmuch_message_get_tags(msg);
```
gauteh: Is this struct safe when message is removed?

src/modes/edit_message.cc
```diff
@@ -666,7 +668,56 @@ namespace Astroid {
...
+    if (draft_msg) {
+      LOG(debug) << "em: set_from called for saved draft";
+       if (a != accounts->get_account_for_address (draft_msg->sender)) {
+         LOG(debug) << "em: set_from called for saved draft with different account";
+
+         ustring fname = draft_msg->fname;
+         path fpath = path(fname.c_str ());
+         if (is_regular_file (fpath)) {
+           boost::filesystem::remove (fpath);
+            ComposeMessage * cmsg = make_draft_message ();
```
gauteh: This accesses and modifies the database. The appropriate locks must be acquired first, and make sure that no dead-locks can occur.

src/modes/edit_message.cc
c-alpha: Good point. Should better make a copy.

src/modes/edit_message.cc
c-alpha: > This accesses and modifies the database. The appropriate locks must be acquired first, and make sure that no dead-locks can occur.

Most of the code is copied from `EditMessage::delete_draft ()`, `EditMessage::save_draft ()`, and `AddDraftMessage::AddDraftMessage (ustring _f)`. Since it all happens in an `OnMessageAction`, the locking should be taken care of by the `ActionManager` (since `class Action` sets `need_db_rw = true` by default)?

> This one is kind of tricky, but will cause crashes in loading thread-indexes or other weird behavior - it might be better to create some generic function in Db and call them there, rather than using the notmuch api directly from this class.

Everything is fine, for as long as the draft message is not the only message on the thread (e.g. when replying), because the thread remains the same across the message removal/re-addition (as the draft message retains `In-Reply-To:` and `References:`, notmuch adds it to the correct threads). Hence, after message removal/re-addition, the thread currently shown in the UI still exists, and the UI and the DB remain in sync.

The trouble starts, when the draft message is alone on a thread. In this case, notmuch generates a new thread ID when the draft is re-added with a new message ID. This causes the UI and the DB to get out of sync, since the UI still holds references to the no longer existing thread ID. A manual "refresh query" cures the problem, but this ought to be automatic.

This is what I am currently working on.

Ideally, I would need a notmuch API along the lines of `notmuch_thread_add_message (message_id, thread_id)` to manually add the "re-branded" draft message to an existing thread, and then removing the old copy of the draft (effectively keeping the existing thread ID, and thus keeping UI and DB in sync). This would however not fit well with notmuch's mental model, in which notmuch decides on its own what's on a thread and why.

As a gross-hack-work-around™, I could also keep the old message in the DB, create a new one with a `References:` pointer to the old one, and add it to the DB. That way, notmuch would add it to the currently shown thread. Then I could remove the old draft (thread survives since the new message is still on it), and then re-remove the `References:` pointer from the new message. An awful lot of work for the DB for little gain.

I am hence currently investigating how to get the UI to close any views for the removed thread, and to update any draft queries.

gauteh: If you emit a thread-updated/thread-changed/message-changed (see source-comments for explanations) on the old thread-id/message-id _after_ deletion, does that not cause the old thread to be removed?

c-alpha: `OnMessageAction` automatically emits after `doit` is complete. If I pass the thread id of the old message to `OnMessageAction`, astroid crashes with a glibmm error:
```
(<unknown>:72931): glibmm-ERROR **:
what: db: could not find thread!
```
Hence I am passing `""` (empty string) as the thread id to `OnMessageAction`.

Here is what I do/happens:
1. do a search for `tag:draft`
1. open a thread view for one of the drafts by arrowing down, and hitting enter (shows "This message is a draft, edit it with E or delete with D." in a red bar at the top)
1. hit `E` to edit the draft
1. result is a "New message" tab with blue bar at the top ("Editing..")
1. change the from address via drop-down
1. my hacked `EditMessage::set_from ()` method is invoked, and the notmuch message is exchanged for a new one in an `OnMessageAction`; since the old draft was the only message on the thread, notmuch also generates a new thread with a new thread id

Now, the "New message" tab is still shown, but the message it displays does no longer exist (it has been deleted in `EditMessage::set_from ()`). The view has become an orphan in a sense. With the "New message" tab still shown, an end user's intuitive next thing to do would be to press `s` to save it. If you do, it seems to work from the UI perspective, but looking at the messages in the terminal reveals that it didn't work, since the underlying `EditMessage` has an empty `fname`. Overall, astroid continues to operate as expected, however, since the new message has been saved and added to the notmuch DB as part of my hacked `EditMessage::set_from ()` method before.

What I would need at this point in the process, would be to get the view of the "New message" tab with blue bar at the top ("Editing..") to reload its contents from the new message file I saved in my hacked `EditMessage::set_from ()` method.

*(NOTE: To continue using astroid, you would have to close the "New message" tab without saving, and then refreshing the `tag:draft` search (hit `$`). END NOTE)*

c-alpha:
On a - somewhat loosely - related note:

`OnMessageAction::emit ()` looks like this:
```c++
  void OnMessageAction::emit (Db * db) {
    astroid->actions->emit_message_updated (db, msg_id);

    if (tid != "")
      astroid->actions->emit_thread_updated (db, tid);
  }
```
`ActionManager::emit_message_updated ()` in turn:
```c++
  void ActionManager::emit_message_updated (Db * db, ustring message_id) {
    LOG (info) << "actions: emitted updated signal for message: " << message_id;
    m_signal_message_updated.emit (db, message_id);

    db->on_message (message_id,
        [&] (notmuch_message_t * nm_msg) {
          if (nm_msg != NULL) {
            const char * tidc = notmuch_message_get_thread_id (nm_msg);
            ustring tid;
            if (tidc != NULL) {
              tid = ustring(tidc);

              emit_thread_changed (db, tid);

            }
          }
      });
  }
```
Doesn't this emit **twice** for the given thread?

gauteh:
c-alpha writes on mars 13, 2018 21:54:
> Doesn't this emit **twice** for the given thread?
> 

Almost! This is confusing, but `emit_thread_changed` and 
`emit_thread_updated` is **not** the same. Please check out the source 
comments for the two for the differences. I have made two similar 
signals in order to optimize event handling between thread-index and 
thread-view where not always an extra update of all the messages is 
needed, but just the thread meta-data.

src/modes/edit_message.cc
gauteh: Yes, OnMessageAction should lock the Db correctly - sorry, did not realize it was wrapped in that one!

c-alpha:
I **believe** to have fixed the WebKitDOM and GLib-GObject messages, as well as the unexpected behaviour with commit e9b8369bbbde76fc338c335a47f3299c74e23a92. Keeping fingers crossed... 🙏

src/utils/ustring_utils.cc
```diff
@@ -53,19 +54,39 @@ namespace Astroid {
...
   }
+  
+  ustring UstringUtils::random_alphanumeric (int length) {
+    const ustring _chars = "abcdefghijklmnopqrstuvwxyz1234567890";
+    return UstringUtils::random_string_from_charset(_chars, length);
+  }
+
+  ustring UstringUtils::random_rfc5322_atext (int length) {
+    // 81 allowed charaters for the message id as per RFC 5322
+    const ustring _chars = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz1234567890!#$%&'*+-/=?^_`{|}~";
```
gauteh: I still think we should remove these: `!#$%&'*+-/=?^_`{|}~` :)

c-alpha: No worries; t'is on my list and will be coming with one of the next commits.

gauteh:
I think this PR does not apply to the current master, and probably still have issues related to message-id and filename discrepancies. Is this something that you want to pursue further or should this PR be closed?

gauteh:
Alexander Adolf writes on September 28, 2018 13:23:
>> I think this PR does not apply to the current master, and probably still have issues related to message-id and filename discrepancies.
> 
> Under the hood, all works fine. The filenames are decoupled from the message-id, and the notmuch DB is updated accordingly. What I got stuck on is updating the UI after the message-id has been changed. I would need to find all views which show messages that have the draft tag set, and then cause those views to reload/rebuild from the notmuch DB.
> 
>> Is this something that you want to pursue further or should this PR be closed?
> 
> I had similar thoughts the other day. The current HEAD has probably evolved too much to still allow this PR to be relevant. I am hence closing this PR. I'll port the changes to the current HEAD once I will have succeeded getting the latest webkigtk working on Mac (need this ti be able to build&test astroid), and will submit a new PR in the future.
> 

Sounds good. Newest webkit 2.22 has some issues as well (#550), but the 
previous one works.

c-alpha:
> Sounds good. Newest webkit 2.22 has some issues as well (#550), but the
> previous one works.

2.22.1 or 2.21.92?

gauteh:
Alexander Adolf writes on September 28, 2018 14:09:
>> Sounds good. Newest webkit 2.22 has some issues as well (#550), but the
>> previous one works.
> 
> 2.22.1 or 2.21.92?

Not sure.

c-alpha:
On 2018-09-28, at 14:09 , Gaute Hope <notifications@github.com> wrote:

> [...]
> > 2.22.1 or 2.21.92?
> 
> Not sure.

Well, I guess I would need to know for which one to make my Homebrew formula...?

gauteh:
Alexander Adolf writes on September 28, 2018 14:16:
> On 2018-09-28, at 14:09 , Gaute Hope <notifications@github.com> wrote:
> 
>> [...]
>> > 2.22.1 or 2.21.92?
>> 
>> Not sure.
> 
> Well, I guess I would need to know for which one to make my Homebrew formula...?

I have not tested myself. See the latest answers in #550. I think I saw 
someone mention 2.21 working.

PR #458, opened 2018-01-26
Fix passing flags to compile tests

rhn:
`try_run`, used by `CheckCXXSourceRuns`, will get confused and pass flags to cmake instead if they are connected with ";", resulting in the error:

```
-- Performing Test GMIME_MATCHES
Parse error in command line argument: -D_LARGEFILE64_SOURCE
Should be: VAR:type=value
CMake Error: No cmake script provided.
CMake Error at /usr/share/cmake/Modules/CheckCXXSourceRuns.cmake:85 (try_run):
  Failed to configure test project build system.
Call Stack (most recent call first):
  CMakeLists.txt:217 (CHECK_CXX_SOURCE_RUNS)
```

This patch solves it by using the safer "," separator for flags.


gauteh:
Thanks, merged.

Issue #461, opened 2018-02-04
Preview mode: replace 'Enter' with 'E' for editing a (draft) email


hugoroy:
There are two ways to edit an draft email in Astroid:

 1. In the thread view, when there is a draft in a thread (with a draft tag), you can edit with 'E'
 2. in the edit view, when you just started a draft email, you can edit with 'Enter'.

This is not consistent. Moreover, 'Enter' is already used in many other contexts (e.g. opening a selected attachment, displaying an HTML part, etc.). This may result in confusing situations (e.g. trying to preview an attachment in the Edit view is counterintuitive because the user has to remember to do it in a different way even though it is visually very similar from the thread view).

I propose to make a consistent approach by using 'e' in any contest to edit an email (whether a mail tagged as draft, or wether a mail in edit mode).

For those who prefer the old baheviour, they can just add the following to their `keybindings` file:

```
thread_view.edit_draft=E
edit_message.edit=GDK_KEY_Return
```

kirschner:
I like that suggestion! I have actually now manually changed it for myself, but think that if it would be the default it would help new users a lot.

PR #462, opened 2018-02-04
consistent editing see issue #461

hugoroy:


gauteh:
Does this mean that we can drop the C+space binding in edit mode?

hugoroy:
This is for attachments, right?

I think so, yes :-)

gauteh:
Hugo Roy writes on februar 5, 2018 16:23:
> This is for attachments, right?
> 
> I think so, yes :-)
> 

Yes, and the HTML part when composing markdown messages - or forwarding 
mime-messages.

gauteh:
So I think we need:

* `e` is always edit
* `E` is always expand (not implemented yet)
* `Enter` is always activate (open or expand)

Am I missing some functionality now?

mxmehl:
`E` is also used to switch the encryption in new message view.

gauteh:
Haven't forgotten this, but super-busy atm so will look into this in time. Still needs some clarification I think (re @mxmehl comment) and updates.

Issue #446, opened 2017-12-19
Remove "format=flowed"


mxmehl:
Astroid sends email with the header `Content-Type: text/plain; charset=utf-8; format=flowed`. The last part seems to create some embarrassing line breaks, especially when writing longer lines than 76 encoded characters (or umlauts which stretch the editor's wrap width after encoding).

Anyway, `format=flowed` [seems to be a rather broken standard](https://blog.fastmail.com/2016/12/17/format-flowed/), and most mail clients don't use it (any more). Is there any reason astroid uses it? And would it cause issues when removing it?

mrvdb:
Please keep it as an option

kirschner:
I also had quite embarrassing problems with that, as the e-mail looked absolutely unprofessional because of that bug. Thanks @mxmehl for researching and reporting!

mxmehl:
Perhaps as additional info: The bug only seems to occur on mail clients which support format=flowed. In my and @kirschner's case this was on Claws Mails IMHO.

kirschner:
First time I got that feedback was from an Evolution user, though.

mrvdb:
@mxmehl Are there more details on the bug other than "embarassing line breaks" ? I've been a long time claws user and migrated to astroid from it.

mxmehl:
I don't have a screenshot to share but I'll try to explain.

In vim I have a fixed line width of 72. In quoted-printable, the maximum width is 76 characters. For example, if you use special characters like umlauts which are escaped by quoted-printable you exceed the 76 character limit. A raw text therefore could look like this (3 lines):

```
Bj=C3=B6rn, k=C3=B6nntest Du mir bitte die besagte Mail komplett forwarden?=
 Dann
gucke ich mal, ob man da was herausfinden kann.
```
In my original text, `Dann` was on the first line, and after it came a normal line break. The `=` at the end of the first line indicates a soft line break which shouldn't be visible after decoding quoted printable. So in the end, it should be 2 lines.

On a client which respects format=flowed, this looks like:

```
Björn, könntest Du mir bitte die besagte Mail komplett forwarden?
 Dann
gucke ich mal, ob man da was herausfinden kann.
```

So format=flowed seems to interpret the soft line break as a hard line break. 

And since the `=` is also used as escape character, the wrong interpretation also causes umlauts to break. This happens if an escaped umlaut (e.g. `=C3=BC` for `ü`) is cut in between by a soft line break:

```
[...]

PR #463, opened 2018-02-05
Disable format=flowed as Content-Type for new mails

mxmehl:
This is a very simplistic approach to solve #446. It just does not add `format=flowed` in the Content-Type header.

format=flowed causes some sick issues in combination with quoted-printable and special characters because astroid (or gmime) does no proper space stuffing to mitigate the automatic line breaks caused by quoted-printable. This can not be resolved by changes in the editor AFAIU. More about the problem in the issue.

Some people have raised that they would like to keep format=flowed as an option which isn't covered in this PR because I lack the skills to realise it. Nevertheless this PR solves an existing bug and doesn't really make new mails extremely ugly or something, so we could still work on adding an improved implementation of format=flowed later IMHO (except someone comes up with a solution quite soon).

cc @kirschner


gauteh:
Thanks, merged, I added an config option (`mail.format_flowed`) default true.

PR #474, opened 2018-02-19
Add Header Files to Xcode Projects

c-alpha:
Xcode needs to be told explicitly about the header files to show them in
the project, and to be able to automatically select them in the\
assistant editors.


gauteh:
Merged.

PR #478, opened 2018-03-06
em: Close page automatically after successful send

mrvdb:
Add configure option: `mail.close_on_success`

### DIFF
--- a/History.txt
+++ b/History.txt
@@ -1,5 +1,5 @@
 == **
-
+* `mail.close_on_success` configures if page closes on successful send.
 * `terminal.font_description` configures font of embedded terminal.
 * `termina.height` configures height of embedded terminal.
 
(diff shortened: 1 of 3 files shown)


mxmehl:
Cool idea! Would it be possible to add a short wait time (2-5 secs for example) and the possibility to skip auto-closing during this time, e.g. by pressing Ctrl-C?

mrvdb:
I'm sure it's technically possible. (sounds pretty similar to the send delay already there)

gauteh:
Nice, maybe we can also add a feature where closing the window while the message is sending (C-w) causes the close_on_successful_send to be set and an small info to be set that it will be closed on sent, and then cycle to the tab which would otherwise get focus if it was closed? 

I have one small concern, if you have accounts where the sent message is not stored (and the email service does not do it for you) you will not have a chance to save the sent message. But I guess that would probably be overly protective..

Either way; I think this can be merged as-is and then the other features can be added later.

gauteh:
Merged.

gauteh:
By the way; please add a section to the configuration reference.

mrvdb:
> By the way; please add a section to the configuration reference.

Push was already waiting, it should be there

gauteh:
Marcel van der Boom writes on mars 6, 2018 18:38:
>> By the way; please add a section to the configuration reference.
> 
> Push was already waiting, it should be there

Great!

PR #482, opened 2018-03-17
fall back to external editor when running in wayland session

vhdirk:
I sync my dotfiles between 2 machines. One is running wayland, the other X11.
Since https://bugzilla.gnome.org/show_bug.cgi?id=721224 still seems to be open, there is no alternative other than to fall back to the external editor.

### DIFF
--- a/src/modes/edit_message.cc
+++ b/src/modes/edit_message.cc
@@ -95,6 +95,12 @@ namespace Astroid {
 
 # ifndef DISABLE_EMBEDDED
     embed_editor = !editor_config.get<bool> ("external_editor");
+    char const* xdg_session_type = std::getenv("XDG_SESSION_TYPE");
+    LOG (info) << "xdg_session_type: " << xdg_session_type;
+    if(xdg_session_type && std::string(xdg_session_type) == "wayland"){
+      LOG (warn) << "Embedded editor not available under wayland. Falling back to external";
+      embed_editor = false;
+    }
 # endif
     save_draft_on_force_quit = editor_config.get <bool> ("save_draft_on_force_quit");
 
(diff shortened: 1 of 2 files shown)


src/modes/edit_message.cc
```diff
@@ -95,6 +95,12 @@ namespace Astroid {
 
 # ifndef DISABLE_EMBEDDED
     embed_editor = !editor_config.get<bool> ("external_editor");
+    char const* xdg_session_type = std::getenv("XDG_SESSION_TYPE");
+    LOG (info) << "xdg_session_type: " << xdg_session_type;
```
gauteh: Change to (debug) and move into if; add else-condition for xdg_session_type NULL.

src/modes/edit_message.cc
```diff
@@ -95,6 +95,12 @@ namespace Astroid {
...
+    if(xdg_session_type && std::string(xdg_session_type) == "wayland"){
```
gauteh: use space after if, and before {

src/modes/edit_message.cc
```diff
@@ -95,6 +95,12 @@ namespace Astroid {
...
+      LOG (warn) << "Embedded editor not available under wayland. Falling back to external";
```
gauteh: add note about: please check your configuration. Also I think this should be changed to error since the editor line specified by the user now probably is not a valid command.

src/modes/editor/external.cc
```diff
@@ -28,7 +28,7 @@ namespace Astroid {
     editor_started = true;
 
     ustring cmd = ustring::compose (editor_cmd,
-        em->tmpfile_path.c_str ());
+        em->tmpfile_path.c_str (), "''", "-1");
```
gauteh: Why not `''` on both? I assume this works if you have a valid external editor command as well where there are no `%2` and `%3`?

PR #495, opened 2018-05-10
Allow to disable gpg throw-keyids

phlb:
By default astroid enable the gpg throw-keyids option. This is useful
for most people. For those who have multiple gpg keys, this is a pain.

This commit add a configuration option  'crypto.gpg.throw_keyids'
allowing to disable this feature.

### DIFF
--- a/src/config.cc
+++ b/src/config.cc
@@ -268,6 +268,7 @@ namespace Astroid {
     /* crypto */
     default_config.put ("crypto.gpg.path", "gpg2");
     default_config.put ("crypto.gpg.always_trust", true);
+    default_config.put ("crypto.gpg.throw_keyids", true);
 
     /* saved searches */
     default_config.put ("saved_searches.show_on_startup", false);
--- a/src/config.cc
+++ b/src/config.cc
@@ -268,7 +268,6 @@ namespace Astroid {
     /* crypto */
     default_config.put ("crypto.gpg.path", "gpg2");
     default_config.put ("crypto.gpg.always_trust", true);
-    default_config.put ("crypto.gpg.throw_keyids", true);
 
     /* saved searches */
     default_config.put ("saved_searches.show_on_startup", false);
--- a/src/crypto.cc
+++ b/src/crypto.cc
@@ -17,6 +17,7 @@ namespace Astroid {
     config       = astroid->config ("crypto");
     gpgpath      = ustring (config.get<std::string> ("gpg.path"));
     always_trust = config.get<bool> ("gpg.always_trust");
+    throw_keyids = config.get<bool> ("gpg.throw_keyids");
 
     LOG (debug) << "crypto: gpg: " << gpgpath;
 
@@ -239,7 +240,7 @@ namespace Astroid {
         sign,
         userid.c_str (),
 # if (GMIME_MAJOR_VERSION >= 3)
-        GMIME_ENCRYPT_THROW_KEYIDS, // only in Gmime 3
+        (throw_keyids ? GMIME_ENCRYPT_THROW_KEYIDS : GMIME_ENCRYPT_NONE), // only in Gmime 3
 # else
         GMIME_ENCRYPT_NONE,
 # endif
--- a/src/crypto.cc
+++ b/src/crypto.cc
@@ -17,7 +17,6 @@ namespace Astroid {
     config       = astroid->config ("crypto");
     gpgpath      = ustring (config.get<std::string> ("gpg.path"));
     always_trust = config.get<bool> ("gpg.always_trust");
-    throw_keyids = config.get<bool> ("gpg.throw_keyids");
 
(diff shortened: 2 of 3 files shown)


mrvdb:
This would indeed be useful for me, I have a local patch in place which is basically the same as this PR.  

Note: I needed to add it to cater for clients on the receiving end which could not handle the throw-keyids option. (somewhat older thunderbird versions)  

The option is useful as such, but shouldn't astroid just follow the way the option has been set in gnupg?

gauteh:
Marcel van der Boom writes on May 10, 2018 12:50:
> This would indeed be useful for me, I have a local patch in place which is basically the same as this PR.  
> 
> Note: I needed to add it to cater for clients on the receiving end which could not handle the throw-keyids option. (somewhat older thunderbird versions)  
> 
> The option is useful as such, but shouldn't astroid just follow the way the option has been set in gnupg?

Thanks for the PR!

I think gnupg config overrides, but not sure. I remember we discussed 
this @mrvdb, was it on IRC - I couldn't find it. Either way, I think we 
landed on default off + config option. Right?

There is definetely an argument for just leaving this to gnupg config, 
but I think that you often want to specify this specifically for e-mail. 
And also it is an important and probably often overlooked option which 
therefore might be useful to bring out into the light with our own 
config option.

mrvdb:
I do recall the conversation indeed. I'm not sure gnupg overrides though, otherwise I would not have needed the patch I think.

Ideal behaviour for me would be:
- default: follow whatever gnupg would set
- allow to override with option

gauteh:
Marcel van der Boom writes on May 10, 2018 13:11:
> I do recall the conversation indeed. I'm not sure gnupg overrides though, otherwise I would not have needed the patch I think.

There are two flags for gpg: --throw-keyids and --no-throw-keyids, so 
that probably depends on whether gmime only adds the first or also the 
second when throw-keyids are not specified. Should be easy enough to 
test.
 
> Ideal behaviour for me would be:
> - default: follow whatever gnupg would set
> - allow to override with option

Sounds good to me. This might be the case depending on GMime behavior 
above, since GMime/astroid option might only override on, but not 
override off. If not the case, I think we should go for the simpler 
solution for now as suggeted in this PR (except maybe default off). 
Though I would definitely be interested in an extended PR.

phlb:
I've proposed this patch because just adding "no-throw-keyids" in my gnupg 
config was not sufficient.


I see 2 propositions that sounds good to me :

    - allow users to force throw-keyids or follow the gpg config. 
      (In that case, perhaps the option should be called 
      "crypt.gpg.force_throw_keyids")

    - just follow the gpg config.

I think that following the gpg config is the best solution (simpler, 
only one place to configure it). But this change the default astroid 
behavior and I don't know if gmime 3 follow the gnupg config for this 
option. Anyone knows ?



Excerpts from Marcel van der Boom's message of mai 10, 2018 12:50 pm:

phlb:
Excerpts from Gaute Hope's message of mai 11, 2018 7:23 am:
> phlb writes on May 10, 2018 15:10:
>> I've proposed this patch because just adding "no-throw-keyids" in my gnupg 
>> config was not sufficient.
>> 
>> 
>> I see 2 propositions that sounds good to me :
>> 
>>     - allow users to force throw-keyids or follow the gpg config. 
>>       (In that case, perhaps the option should be called 
>>       "crypt.gpg.force_throw_keyids")
>> 
>>     - just follow the gpg config.
>> 
>> I think that following the gpg config is the best solution (simpler, 
>> only one place to configure it). But this change the default astroid 
>> behavior and I don't know if gmime 3 follow the gnupg config for this 
>> option. Anyone knows ?
> 
> GMime does not, at least not in the absence of a `throw-keyids` option. 
> Otherwise, there would not any point in the flag we can pass from 
> astroid. Since you already tested with `no-throw-keyids` it seems like 
> it does not respect that either. It would be interesting to know what 
> happens when disabled in GMime, and enabled in gnupg config?
> 
> If still off: we need our own option, in which case your patch as it is 
> is good.
> 
> If on: we can remove the throw-keyids stuff from astroid and gnupg 
> config will take precedence. This option/behavior is not very old, and it has 
> been controversial, so I think it is fine to change the default astroid 
> behavior.

Gmime follow the throw-keyids option in the gpg configuration when 
disabled in Gmime. Tested with gmime 3.2.0 and gpg 2.2.5

So I've updated this patch to simply remove forcing this option in 
astroid code.

Philippe.

gauteh:
phlb writes on May 11, 2018 11:01:
> Gmime follow the throw-keyids option in the gpg configuration when 
> disabled in Gmime. Tested with gmime 3.2.0 and gpg 2.2.5
> 
> So I've updated this patch to simply remove forcing this option in 
> astroid code.
> 

Thanks, squashed and merged. Ideally we should warn the user when Bcc's 
are encrypted to and that the information is leaking. I'll open an issue 
about that.

PR #513, opened 2018-07-03
cf: do not write back config

### LINKED ISSUE #511: astroid crashes trying to update non-writable config
Astroid attempts to add missing values to the config on startup. When the config is non-writable, such as in my case (specifically, because it is managed by [home manager](https://github.com/rycee/home-manager)), this crashes. It makes it a pain to update astroid in general, as I then have to manually a) make it writable, b) figure out the changes, c) integrate the changes with my config template.

Failing gracefully would be appreciated, not just because it would make updating easier for me, but it would also make my config cleaner, containing only the things I've actually *changed*.

gauteh:
fix #511


src/account_manager.cc
```diff
@@ -28,33 +28,74 @@ namespace Astroid {
...
+      a->isdefault =
+        (bv = kv.second.get_optional<bool> ("default"),
+        (bv ? bv.get () : false));
+
+      a->save_sent =
+        (bv = kv.second.get_optional<bool> ("save_sent"),
+        (bv ? bv.get () : false));
+
+      a->save_sent_to = Utils::expand(bfs::path (
+            (sv = kv.second.get<string> ("save_sent_to"),
```
tkerber: `get_optional` is needed here, line 76, and 81.

PR #517, opened 2018-07-05
take DESTDIR environment variable into account (if and when set) when installing symlinks

c-alpha:
This should address astroidmail/astroid#514.

I myself cannot test at this time because of astroidmail/homebrew-astroid#11 (no WebKitGTK+ on macOS at the moment). **So please do let me know whether this works and solves the issue.**


gauteh:
Thanks, merged! :)

Hello71:
I don't get why it has to check DESTDIR. `//a` is the same as `/a` anyways (for the source of symlink).

gauteh:
`/a/b/` is not the same as `a/b/`

Hello71:
still not sure it works right, but I don't know how cmake install works, so I'll assume it's right and report back if it breaks for me.

PR #522, opened 2018-07-10
support BUILD_TESTING cmake variable

Hello71:
note that I didn't actually test this (as one might expect :P)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -534,6 +534,7 @@ endif()
 ##
 # Tests
 #
-enable_testing()
-add_subdirectory (tests)
-
+if(BUILD_TESTING)
+  enable_testing()
+  add_subdirectory (tests)
+endif()


gauteh:
Alex Xu writes on July 10, 2018 18:40:
> note that I didn't actually test this (as one might expect :P)

Thanks. Hehe ;) You shouldn't say that. Anyway, what is the use of this 
option? Is this a common option? You can build only astroid by doing 
e.g.: $ ninja astroid (or use the astroid target for your build system).

I don't think it works as-is, you probably need to add the var here: https://github.com/Hello71/astroid/blob/4eb36e1a291954c8cd4c47d07d9e5fbc8012b02a/CMakeLists.txt#L63

Needs testing!

Hello71:
I tested BUILD_TESTING=NO and it works as is. Your solution probably works too, but I don't want to put it into a distro package on the basis that you may want to add more programs later.

gauteh:
Alex Xu writes on July 10, 2018 20:52:
> I tested BUILD_TESTING=NO and it works as is. Your solution probably 
> works too, but I don't want to put it into a distro package on the 
> basis that you may want to add more programs later.
> 

I'm not convinced about these changes. Tests should be built by default, 
and there are ways to build only the astroid target if desired. I am 
neither familiar with this option as common to projects - or expected by 
various package systems.

Hello71:
Gentoo supports a "test" USE flag. While support is not mandatory, most users don't want to run tests for every package on their machine, on the basis that someone probably already has. I picked BUILD_TESTING since CTest already supports it, although from a Gentoo perspective we don't really care.

gauteh:
Alex Xu writes on July 11, 2018 16:30:
> Gentoo supports a "test" USE flag. While support is not mandatory, most users don't want to run tests for every package on their machine, on the basis that someone probably already has. I picked BUILD_TESTING since CTest already supports it, although from a Gentoo perspective we don't really care.
> 

Is this the option you refer to:  https://cmake.org/cmake/help/latest/module/CTest.html ? Is it always ON by default?

In that case should we also do `include (CTest)`? 

[ It would be useful with some rationale and maybe a link to that option in 
  the commit message ]

@c-aplha: would you care to review this as well - you're better tuned to 
CMake / Astroid ;)

Hello71:
Ah, I thought it already had `include(CTest)`.

c-alpha:
I don't think `BUILD_TESTING` does what is intended here. See docs [here](https://cmake.org/cmake/help/v3.12/module/CTest.html?highlight=build_testing) and [here](https://cmake.org/cmake/help/v3.12/command/add_test.html?highlight=build_testing).

I would suggest to instead declare a new `option()` (e.g. `ENABLE_TESTS`) which is used along the lines of what is proposed in the code changes of this PR, and which defaults to `OFF` for release builds, and to `ON` for debug builds.

Packagers and distros (such as e.g. Homebrew) should then build release versions (`-DCMAKE_BUILD_TYPE=Release`). Since debug is the default build type (at least for now), developers will still see the same behaviour as currently.

And with the new option, if anyone insists they can still build a release version with the tests, too (`-DCMAKE_BUILD_TYPE=Release -DENABLE_TESTS=ON`).

gauteh:
Alexander Adolf writes on August 29, 2018 22:52:
> Packagers and distros (such as e.g. Homebrew) should then build release versions (`-DCMAKE_BUILD_TYPE=Release`). Since debug is the default build type (at least for now), developers will still see the same behaviour as currently.
> 
> And with the new option, if anyone insists they can still build a release version with the tests, too (`-DCMAKE_BUILD_TYPE=Release -DENABLE_TESTS=ON`).

Most distros build and run tests, so I think tests should always be 
built by default.

c-alpha:
> [...]
> Most distros build and run tests, so I think tests should always be 
> built by default.

Ok, so we'll have tests on by default. But we should have an explicit
`option()` in the `CMakeLists.txt` instead of the current proposal.

gauteh:
Closing, possible through other methods (make astroid) and need changes.

Issue #397, opened 2017-09-07
astroid manpage


teto:
Trying `man astroid` in nixos but doesn't work. As nixos derivations often forget to install man pages, I tried to fix it but can't find the man in the source so here is my question; does a manpage exist and if yes, where is it please ?

gauteh:
Matthieu Coudron writes on september 7, 2017 6:27:
> Trying `man astroid` in nixos but doesn't work. As nixos derivations often forget to install man pages, I tried to fix it but can't find the man in the source so here is my question; does a manpage exist and if yes, where is it please ?

It does not exist. Yet.

gauteh:
Fnux writes on July 24, 2018 9:26:
> I'm willing to help! Is it okay for you if I use [scdoc](https://drewdevault.com/2018/05/13/scdoc.html) to generate the man pages ? You can find an example [here](https://github.com/swaywm/sway/blob/master/sway/sway.1.scd).
> 

Great! That would be really good!

scdoc looks promising, but I am a bit concerned about its availabilitiy:
I just checked with Arch Linux and it is only available in AUR. How 
is its availabilitiy in other distros? If it is not available then it 
probably won't help on e.g. Debian?

Fnux:
`scdoc` is already available (well, I packaged it) on Fedora: https://apps.fedoraproject.org/packages/scdoc

In order to be nice to distributions that do not package `scdoc` (yet), we can ship pre-generated manpages with the release archives. Manual pages are not considered as "precompiled binaries" :-)

Moreover, `scdoc` is super-easy to package: a simple C program without any external dependency.

gauteh:
Fixed in #530.

PR #530, opened 2018-07-30
Manual page: astroid(1)

Fnux:
As discussed in #397, here is a simple manpage for astroid. Most of the content is extracted from the wiki.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,6 +47,11 @@ $ cd build
 $ ctest
 ```
 
+#### Manual page
+
+We use [scdoc](https://drewdevault.com/2018/05/13/scdoc.html) to generate
+`astroid(1)`:  `scdoc < doc/astroid.1.scd > doc/astroid.1`.
+
 ### Installing
 
 Configure with a prefix and install:
--- a/doc/astroid.1.scd
+++ b/doc/astroid.1.scd
@@ -28,6 +28,10 @@ tons of e-mail.
 *-h, --help*
 	Show the help message and quit.
 
+*--log-level <level>*
+	Override the configured log level. Level options are: trace, debug (default),
+	info, warn, error and fatal.
+
 *--log-stdout*
 	Output logs to the standard output, regardless of any other logging
 	configuration.
--- a/doc/astroid.1.scd
+++ b/doc/astroid.1.scd
@@ -1,0 +1,107 @@
+astroid(1)
+
+# NAME
+
+astroid - graphical threads-with-tags style, lightweight and fast, e-mail
+client for Notmuch
+
+# SYNOPSIS
+
+*astroid* [options]
+
+# DESCRIPTION
+
+Astroid is a lightweight and fast *Mail User Agent* that provides a graphical
+interface to searching, displaying and composing email, organized in threads
+and tags. It uses the Notmuch backend for blazingly fast searches through
+tons of e-mail.
+
+# OPTIONS
+
+*-c, --config* <config>
+	Override the default configuration path. Please refer to the *CONFIGURATION*
+	section for details.
+
+*--disable-log*
+	Disable logging.
+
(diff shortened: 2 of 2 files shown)


gauteh:
Thanks, this looks very good! 

I'll see if I can make a processor for CMake, since scdoc is not very 
widely available this will gracefully fail if scdoc is missing. I think 
we should drop or change the man page section in the README to just say along with the note about the wiki: ... or see `man astroid` (requires `scdoc`).

A few comments:

* Notmuch is required (the way it is written now is seems like it is 
  optional).
* Perhaps also refer to gmailieer or to the notmuch wiki in this list
* The environmental variable: ASTROID_DIR can be used to set the path of 
  the local resources. This is mostly for debug purposes or when running 
  directly from source (see examples/astroid). Not sure if this should 
  be documented in the man page.
* The environmental variable: GI_TYPELIB_PATH should be set to the 
  typelib dir when astroid is not installed to be able to use plugins. 
  Also, not sure if this should be included. It is more of a libpeas 
  thing / gobject-introspection thing.
* The command line option: --log-level is in master which can be used to 
  override the configured log level (options are trace,debug (default),info,warn,error,fatal)

Fnux:
> * The environmental variable: ASTROID_DIR can be used to set the path of
  the local resources. This is mostly for debug purposes or when running
  directly from source (see examples/astroid). Not sure if this should
  be documented in the man page.

What "local resources" are you talking about ? Configuration ?

> * The environmental variable: GI_TYPELIB_PATH should be set to the
  typelib dir when astroid is not installed to be able to use plugins.
  Also, not sure if this should be included. It is more of a libpeas
  thing / gobject-introspection thing.

I don't think it is useful to the end user, and would be more confusing than helping.

gauteh:
Fnux writes on August 8, 2018 13:12:
>> * The environmental variable: ASTROID_DIR can be used to set the path of
>   the local resources. This is mostly for debug purposes or when running
>   directly from source (see examples/astroid). Not sure if this should
>   be documented in the man page.
> 
> What "local resources" are you talking about ? Configuration ?

Theme files (html, scss), webextension, glade files. These are normally 
placed in ~/.config/astroid when user-configured (unrelated to this 
var). So this var can be undocumented.

> 
>> * The environmental variable: GI_TYPELIB_PATH should be set to the
>   typelib dir when astroid is not installed to be able to use plugins.
>   Also, not sure if this should be included. It is more of a libpeas
>   thing / gobject-introspection thing.
> 
> I don't think it is useful to the end user, and would be more confusing than helping.

Agreed.

The page you have written can also be used with `ronn` which is more 
widely available. I think we can use this in the Makefiles if scdoc is 
not avilable.

Fnux:
> Theme files (html, scss), webextension, glade files. These are normally
placed in ~/.config/astroid when user-configured (unrelated to this
var). So this var can be undocumented.

I feel like those files should go in `$XDG_DATA_HOME/astroid` instead of `$XDG_CONFIG_HOME/astroid`. Some user might use the `ASTROID_DIR` environment variable, I will add it to the manpage.

Fnux:
I misunderstood, my bad. Let's ignore ASTROID_DIR.

gauteh:
Fnux writes on August 24, 2018 11:30:
> I misunderstood, my bad. Let's ignore ASTROID_DIR.

Sounds good. Are there any more changes you plan on doing before the PR 
is ready? I probably won't be able to merge it before next week.

Fnux:
That's all :-)

gauteh:
Fnux writes on August 24, 2018 11:40:
> That's all :-)
> 

Merged! The manpage is now built and installed. If scdoc is not 
available ronn is used. I also documented the `$NOTMUCH_CONFIG` variable 
which is now supported (#543, @kyrias).

PR #531, opened 2018-07-30
Problem: tests/run-test.sh might fail

yrashk:
This is because on some systems, bash is not in
/bin/bash

Solution: use /usr/bin/env bash in the shebang


gauteh:
Thanks, merged!

PR #537, opened 2018-08-24
cmake: fix scalable icon install path

TristanCacqueray:
The correct path is apps, not app*.


gauteh:
Tristan de Cacqueray writes on August 24, 2018 5:17:
> The correct path is apps, not app*.

Thanks, merged.

PR #538, opened 2018-08-24
cmake: libhypocycloid should always be static

gauteh:


TristanCacqueray:
Perhaps mention in the commit message that fix an issue when cmake is executed with -DBUILD_SHARED_LIBS:BOOL=ON .

Thanks for the fast fix!

gauteh:
Tristan de Cacqueray writes on August 24, 2018 9:46:
> Perhaps mention in the commit message that fix an issue when cmake is executed with -DBUILD_SHARED_LIBS:BOOL=ON .
> 
> Thanks for the fast fix!

Done and merged.

PR #543, opened 2018-08-28
Use NOTMUCH_CONFIG if set, otherwise expand astroid.notmuch_config

kyrias:


gauteh:
Thanks, merged.

gauteh:
Please update the configuration reference in the wiki w.r.t. to the 
config.db option.

Issue #532, opened 2018-07-30
Problem: composing emails when thread_view.preferred_type = html


yrashk:
When `thread_view.preferred_type` is set to `html`, composing a plain text email ends up in wiping out message body after exiting from the editor.

When `preferred_type` is reverted back to `plain`, everything works as expected, but we lose the ability to see emails in the mailbox as HTML by default.

Here's a test that confirms this:

```
  BOOST_AUTO_TEST_CASE (compose_test_body_preferred_html)
  {
    using Astroid::ComposeMessage;
    using Astroid::Message;
    setup ();
    const_cast<ptree&>(astroid->config("thread_view")).put ("preferred_type", "html");

    ComposeMessage * c = new ComposeMessage ();

    ustring bdy = "This is test: æøå.\n > testing\ntesting\n...";

    LOG (trace) << "cm: writing utf-8 text to message body: " << bdy;
    c->body << bdy;

    c->build ();
    c->finalize ();
    ustring fn = c->write_tmp ();

    delete c;

    Message m (fn);

    ustring rbdy = m.viewable_text (false);

    BOOST_CHECK_MESSAGE (bdy == rbdy, "message reading produces the same output as compose message input");

    unlink (fn.c_str ());

    teardown ();

  }
```

This behaviour happens in Chunk where Chunk queries the configuration and if the type is preferred it sets the `preferred` property on a chunk to `true`. This, in turn, is used in `Message` which is in turn used by `ComposeMessage`. Since no chunk is marked as preferred, none is used.

Proposed solution: if no chunks is found to be preferred, pick one.

P.S. I am trying to produce a patch to fix this, but this might take a bit of time (the fact that I never really used C++ is not helping me!)

gauteh:
Hi, thanks! Really good to have a test-case. Would be awesome with a 
patch! 

Yurii Rashkovskii writes on July 30, 2018 20:42:
> This behaviour happens in Chunk where Chunk queries the configuration and if the type is preferred it sets the `preferred` property on a chunk to `true`. This, in turn, is used in `Message` which is in turn used by `ComposeMessage`. Since no chunk is marked as preferred, none is used.
> 
> Proposed solution: if no chunks is found to be preferred, pick one.
> 
> P.S. I am trying to produce a patch to fix this, but this might take a bit of time (the fact that I never really used C++ is not helping me!)

Similar to what you suggest I think the issue is in 
`Message::viewable_text (...)`. There is no HTML part and it tries to 
[...]

PR #545, opened 2018-08-30
Preferred html always

gauteh:
#532 

@yrashk I've added your test case in your name. 

* no_newline test fails
* Still some preview problems?


yrashk:
@gauteh Gaute, I was mostly travelling September. I apologize for neglecting this issue. I will return to it shortly.

Issue #501, opened 2018-05-22
SIGSEGV in Astroid::ThreadIndex::on_stats_ready


iain-buclaw-sociomantic:
To reproduce:
- Start astroid.
- `Ctrl+o` to open new window.
- Close new window.
- Press `N` to mark any thread as read.

This happens consistently every time I try this sequence of signal/events.

```
Thread 1 "astroid" received signal SIGSEGV, Segmentation fault.
0x000055555588cf9d in Astroid::ThreadIndex::on_stats_ready (this=0x7ffff454c9a0 <g_nullify_pointer>) at ../src/modes/thread_index/thread_
index.cc:202
(gdb) bt
#0  0x000055555588cf9d in Astroid::ThreadIndex::on_stats_ready() (this=0x7ffff454c9a0 <g_nullify_pointer>) at ../src/modes/thread_index/t
hread_index.cc:202
#1  0x000055555595f49b in Astroid::QueryLoader::on_thread_changed(Astroid::Db*, Glib::ustring) (this=0x55555626ce10, db=0x7fffffffd060, t
hread_id=...) at ../src/modes/thread_index/query_loader.cc:375
#2  0x0000555555961eaf in sigc::bound_mem_functor2<void, Astroid::QueryLoader, Astroid::Db*, Glib::ustring>::operator()(Astroid::Db* cons
t&, Glib::ustring const&) const (this=0x55555634c288, _A_a1=@0x7fffffffcf10: 0x7fffffffd060, _A_a2=...) at /usr/include/sigc++-2.0/sigc++
/functors/mem_fun.h:2143
#3  0x00005555559618e9 in sigc::adaptor_functor<sigc::bound_mem_functor2<void, Astroid::QueryLoader, Astroid::Db*, Glib::ustring> >::oper
ator()<Astroid::Db* const&, Glib::ustring const&>(Astroid::Db* const&, Glib::ustring const&) const (this=0x55555634c280, _A_arg1=@0x7ffff
fffcf10: 0x7fffffffd060, _A_arg2=...) at /usr/include/sigc++-2.0/sigc++/adaptors/adaptor_trait.h:108
#4  0x0000555555960f7f in sigc::internal::slot_call<sigc::bound_mem_functor2<void, Astroid::QueryLoader, Astroid::Db*, Glib::ustring>, vo
[...]

PR #555, opened 2018-09-25
ql: clear actions sigc connections

gauteh:
#501


iain-buclaw-sociomantic:
#558 is the correct fix here.

gauteh:
closed in favor of #558.

PR #560, opened 2018-09-30
fix #550: use srcdoc attribute when setting message content in iframe

### LINKED ISSUE #550: loading threads freezes astroid with webkit2gtk 2.22.0

gauteh:
using the src attribute causes a request that needs to be allowed on the
main GUI thread on new webkit. even when setting the iframe src async it
needs to run on the extension GUI thread in order to be able to
manipulate the DOM tree. thus subsequent calls to the webext, also
necessarily running on the extension GUI thread, blocks .. and the main
GUI blocks while waiting for these to return. this causes a deadlock
while the webext waits for the request to be handled on the main GUI
thread.


gauteh:
As far as I can see this does not allow message content to break out of the IFRAME: https://stackoverflow.com/questions/50140983/does-setattribute-automatically-escape-html-characters

emdete:
Yes, works here too. Thank you gauteh!

PR #565, opened 2018-10-03
Add more full mailto URI parsing

kyrias:


gauteh:
Hi, I tested with the following invocations and had some issue (might be 
user-error, but perhaps some more graceful failures). These would also 
crash other astroid windows if they were already visible.

$ astroid --no -m eg@gaute.vetsj.com
$ astroid --no -m mailto:eg@gaute.vetsj.com
$ astroid --no -m mailto://eg@gaute.vetsj.com

the first fails with: 
```
[08:26:12] [0x00007f9c17a13a00] [M] [info] astroid: mailto: eg@gaute.vetsj.com

(astroid:30299): libsoup-CRITICAL **: 08:26:12.731: soup_uri_get_scheme: assertion 'uri != NULL' failed

(astroid:30299): glibmm-ERROR **: 08:26:12.731: 
what: basic_string::_M_construct null not valid
```

while the second two fail with:
```
[08:26:58] [0x00007f5aca15ba00] [M] [info] astroid: mailto: mailto:eg@gaute.vetsj.com

(astroid:30426): glibmm-ERROR **: 08:26:58.376: 
what: basic_string::_M_construct null not valid
```

Maybe I am using this wrong? Does it make sense to have a switches for 
-s subject, -t to or similar? Since writing a mailto: uri on the cmd 
line is maybe a bit unintuitive and is used mostly when invoking astroid 
from a URL on a webpage or e-mail?

On a related note it would be nice to support piping a draft (or 
supplying one as a file maybe). Especially, git format-patch series or 
similar I'd like to edit or see before sending. In this case it would 
actually be sending a whole new thread! Would be interesting to see if 
we could come up with some concept that could work with git there! 
Thoughts for another feature-request I guess..

gauteh:
Johannes Löthberg writes on October 5, 2018 18:26:
> Ah, silly me.  What's funny is that I wrote it correctly in my first draft, but then I tried a few other libraries before writing it using libsoup again.  The first two should work fine now.  Third is accepted, though it technically shouldn't really be, because the `mailto` scheme doesn't have an authority section, just the path.

Seems to work now. One issue: if I already have an astroid instance 
running the --mailto command works (opens a new message in compose win), 
but results in a segfault in the new invocation. 

I haven't tested with subjects, etc, with an instance already running. 

This will probably be the most common use case.

> And yeah, having more appropriate command line flags would be good as well.  Before astroid I was using alot, which has the following flags that I was using to semi-automate sending emails:
> 
> * `--sender`
> * `--to`
> * `--subject`
> * `--template` (template message body to open in the editor)
> * `--tags`
> 
> Though for something like this, I think it would make sense to have a concept of major command modes, like `alot compose` for alot, to group them under.

Agreed, should be a sub-command. Does `--template` take a file argument?

gauteh:
Gaute Hope writes on October 6, 2018 9:53:
> Seems to work now. One issue: if I already have an astroid instance 
> running the --mailto command works (opens a new message in compose win), 
> but results in a segfault in the new invocation. 

```

Starting program: /home/gaute/dev/mail/notm/astroid/build/astroid --no -m mailto:eg@gaute.vetsj.com
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/usr/lib/libthread_db.so.1".
[New Thread 0x7fffdc2fd700 (LWP 3173)]
[New Thread 0x7fffdbafc700 (LWP 3174)]
[2018-10-06 09:50:19.990177] [0x00007fffddd1da00] [info]    welcome to astroid! - v0.14-3-gcda29352
[New Thread 0x7fffdb2ba700 (LWP 3185)]

Thread 1 "astroid" received signal SIGSEGV, Segmentation fault.
0x000055555596e8e3 in std::vector<Astroid::Account, std::allocator<Astroid::Account> >::~vector (this=0x5, __in_chrg=<optimized out>)
    at /usr/include/c++/8.2.0/bits/stl_vector.h:567
567             std::_Destroy(this->_M_impl._M_start, this->_M_impl._M_finish,
(gdb) bt
#0  0x000055555596e8e3 in std::vector<Astroid::Account, std::allocator<Astroid::Account> >::~vector() (this=0x5, __in_chrg=<optimized out>)
    at /usr/include/c++/8.2.0/bits/stl_vector.h:567
#1  0x000055555596d7fc in Astroid::AccountManager::~AccountManager() (this=0x5, __in_chrg=<optimized out>) at ../src/account_manager.cc:168
#2  0x00005555557d1ce9 in Astroid::Astroid::~Astroid() (this=0x555555b4b1d0, __in_chrg=<optimized out>, __vtt_parm=<optimized out>) at ../src/astroid.cc:408
#3  0x00005555557d1e92 in Astroid::Astroid::~Astroid() (this=0x555555b4b1d0, __in_chrg=<optimized out>, __vtt_parm=<optimized out>) at ../src/astroid.cc:421
#4  0x00007ffff1b2d02b in g_datalist_clear () at /usr/lib/libglib-2.0.so.0
#5  0x00007ffff1e2ab05 in g_object_unref () at /usr/lib/libgobject-2.0.so.0
#6  0x00005555557cc1a1 in Glib::RefPtr<Astroid::Astroid>::~RefPtr() (this=0x555555af81e8 <Astroid::astroid>, __in_chrg=<optimized out>)
    at /usr/include/glibmm-2.4/glibmm/refptr.h:273
#7  0x00007fffedd5623c in __run_exit_handlers () at /usr/lib/libc.so.6
#8  0x00007fffedd5636e in  () at /usr/lib/libc.so.6
#9  0x00007fffedd4000a in __libc_start_main () at /usr/lib/libc.so.6
#10 0x00005555557cbdfe in _start ()
(gdb)
```

gauteh:
Should probably move several of the Astroid:: data members to 
unique_ptr.

Issue #571, opened 2018-10-10
Hook to process message before display/reply?


larskotthoff:
One of my email providers inserts a warning into the body of the email if certain conditions are met. This is annoying and completely useless, and worse shows up in the quoted text when I reply.

It would be great to have a hook before astroid reads a mail that allows to pipe it through a script. I could remove the warning as if it never existed this way, without having to change mails on the server.

larskotthoff:
Thanks, I'll have a look at that. It would be great if you could include something that allows this to hook in before a reply.

larskotthoff:
It looks like if I return the plain/text part of a mail from a plugin all the newlines are gone. This happens even if I return the text part verbatim without changes:

      def do_filter_part(self, text, html, mime_type, is_patch):
        if mime_type == 'text/plain':
          return text

Is this a bug or am I missing something?

larskotthoff:
Also, is there a way to modify the headers?

gauteh:
Lars Kotthoff writes on October 12, 2018 0:37:
> It looks like if I return the plain/text part of a mail from a plugin all the newlines are gone. This happens even if I return the text part verbatim without changes:
> 
>       def do_filter_part(self, text, html, mime_type, is_patch):
>         if mime_type == 'text/plain':
>           return text
> 
> Is this a bug or am I missing something?

You have to return HTML. The html argument is the GMime filtered and 
escaped HTML (so it is pretty safe), this is what is usually displayed, 
use this if possible. The html vs text is related line by line (as far 
as I can remember).

You can probably also access GMimes filters via python if you need to 
re-encode later.

The documentation (in the wiki) is probably a bit sparse, feel free to 
elaborate as you go along :)

It might be useful to provide a `filter_message` or another 
`filter_part` with perhaps a context argument. This could be used 
directly in message reading (chunk.cc / messagethread.cc) to pre-process 
the mail.

larskotthoff:
Thanks, I was looking through the code and came to the same conclusion. I'm starting to think that I really want a plugin interface that allows to process the raw message.

gauteh:
Lars Kotthoff writes on October 12, 2018 18:08:
> Thanks, I was looking through the code and came to the same conclusion. I'm starting to think that I really want a plugin interface that allows to process the raw message.
> 

Yes, the current interface is very ad-hoc for what I need. Could 
definetely be made much better organized and comprehensive! 
Contributions very welcome.

- gaute

larskotthoff:
[...]

PR #577, opened 2018-10-17
More flexible plugin API that allows to process entire message

larskotthoff:
See #571. These changes add a plugin API that passes the file name of the message to the `do_process` method of the plugin and expects a GMimeStream back -- this allows for arbitrary preprocessing of the message, including headers.

Example use case: add plain text part to HTML-only messages so that they can be quoted better when replying.

### DIFF
--- a/src/message_thread.cc
+++ b/src/message_thread.cc
@@ -16,6 +16,9 @@
 # include "utils/ustring_utils.hh"
 # include "utils/vector_utils.hh"
 # include "actions/action_manager.hh"
+# ifndef DISABLE_PLUGINS
+  # include "plugin/manager.hh"
+# endif
 
 using namespace std;
 using namespace boost::filesystem;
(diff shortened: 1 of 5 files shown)


src/message_thread.cc
```diff
@@ -170,30 +173,35 @@ namespace Astroid {
...
+          missing_content = true;
 
-        if (in_notmuch) {
-          LOG (warn) << "loading cache for missing file from notmuch";
-          load_notmuch_cache ();
-        } else {
-          LOG (error) << "tried to open disk file, but failed, message is not in database either.";
-          string error_s = "failed to open file: " + fname;
-          throw message_error (error_s.c_str());
+          if (in_notmuch) {
```
gauteh: The cache loading from notmuch now depends on what the plugin does and wether it manages to open the file. What happens if there's an error in parsing the file? Or if the file does not exist?

larskotthoff: In either case the plugin should return a NULL pointer (unless it's doing some reading from a cache itself) and go into this branch of the code. Technically this will try to open the same file again and fail again which isn't necessary, but it keeps the code simpler.

gauteh:
Lars Kotthoff writes on October 17, 2018 6:11:
> See #571. These changes add a plugin API that passes the file name of the message to the `do_process` method of the plugin and expects a GMimeStream back -- this allows for arbitrary preprocessing of the message, including headers.
> 
> Example use case: add plain text part to HTML-only messages so that they can be quoted better when replying.

Merged. Could you please add the method to the reference in the wiki?

larskotthoff:
Will do; I'll also add an example plugin.

gauteh:
Lars Kotthoff writes on October 18, 2018 17:06:
> Will do; I'll also add an example plugin.

Great! I think we can close the original issue?

larskotthoff:
I've added a short description to the wiki and put an example plugin at https://github.com/larskotthoff/astroid-processing-plugin. Could you add that to the plugins repo please?

gauteh:
Lars Kotthoff writes on October 18, 2018 17:58:
> I've added a short description to the wiki and put an example plugin at https://github.com/larskotthoff/astroid-processing-plugin. Could you add that to the plugins repo please?
> 

Thanks, will do. It would be great if it also works on GMime 3. Would it 
be a big hassle to change that? Some of the other plugins at least used 
to import 3.0 if 2.6 failed (or maybe it was the other way).

larskotthoff:
I'm getting

    gi.RepositoryError: Requiring namespace 'GMime' version '3.0', but '2.6' is already loaded

when I change it to 3.0. Any ideas what's going on there?

gauteh:
Lars Kotthoff writes on October 18, 2018 18:32:
> I'm getting
> 
>     gi.RepositoryError: Requiring namespace 'GMime' version '3.0', but '2.6' is already loaded
> 
> when I change it to 3.0. Any ideas what's going on there?

Hm, weird.. do you have any other plugins running? Maybe they are run in 
the same python instance.

larskotthoff:
No other plugins. At least I'm not running anything explicitly; maybe astroid is loading a plugin that ships with it implicitly?

gauteh:
Lars Kotthoff writes on October 18, 2018 18:38:
> No other plugins. At least I'm not running anything explicitly; maybe astroid is loading a plugin that ships with it implicitly?

There are currently no plugins shipped with astroid.

Issue #317, opened 2017-04-10
Tag mail directly after sending it


mxmehl:
It would be awesome if astroid allowed us to tag a mail directly after (or shortly before) sending it. 

Scenario: I write an email and I want to tag it with something like "waiting-for-reply" or "blocking" so I have a reminder for mails I have to remind people of. 

Currently, this requires me to: write mail > go to sent mails > (open thread > ) tag single email.

Ideally we could tag the mail in the "New Message" window already with the known keybinding. If the mail has been sent, the tags are applied to it.

Would this be possible?

kirschner:
Nice idea. This would also be very helpful for my workflow.

hugoroy:
I expect this could be done already with custom headers and a custom notmuch post new hook. 

(Every email I send gets in my inbox so for me it's just easy to tag it right there).

gauteh:
Hugo Roy writes on juli 5, 2017 14:48:
> I expect this could be done already with custom headers and a custom notmuch post new hook. 
> 

You can configure custom tags to be added automatically already in the 
config, but you can't manually add tags through the UI on the 
'successfully sent message' yet.

hugoroy:
Right.

What happens if you manually add an email header in the text editor and send it to astroid? Will astroid discard the header?

gauteh:
Hugo Roy writes on juli 5, 2017 14:52:
> Right.
> 
> What happens if you manually add an email header in the text editor and send it to astroid? Will astroid discard the header?

At the moment yes, there is a issue regarding that (specifically the 
Reply-To header).

gauteh:
Gaute Hope writes on juli 5, 2017 14:50:
> Hugo Roy writes on juli 5, 2017 14:48:
>> I expect this could be done already with custom headers and a custom notmuch post new hook. 
>> 
> 
> You can configure custom tags to be added automatically already in the 
> config, but you can't manually add tags through the UI on the 
> 'successfully sent message' yet.

This currently only works for messages that are saved directly and added 
to the DB. It is possible to add ghost-messages in notmuch, but we're 
not doing that in astroid at the moment.

mxmehl:
Am 05.07.2017 um 14:55 schrieb Gaute Hope:
>> You can configure custom tags to be added automatically already in the
>> config, but you can't manually add tags through the UI on the
>> 'successfully sent message' yet.
> 
> This currently only works for messages that are saved directly and added
> to the DB. It is possible to add ghost-messages in notmuch, but we're
[...]

PR #581, opened 2018-10-20
Allow to tag sent messages with same tags as message being replied to

larskotthoff:
This PR introduces special values to `accounts.<account>.additional_sent_tags`, namely `*`  (apply all tags that the message being replied to, if any, has), and `-<tag>` (to exclude tags from being applied). For example, the option "foo,*,-attachment" applies the tag "foo" and any tags the message being replied to has, except for "attachment".

### DIFF
--- a/src/db.cc
+++ b/src/db.cc
@@ -327,13 +327,38 @@ namespace Astroid {
     return _mid;
   }
 
-  ustring Db::add_sent_message (ustring fname, vector<ustring> additional_sent_tags) {
+  ustring Db::add_sent_message (ustring fname, vector<ustring> additional_sent_tags, ustring parent_mid) {
     LOG (info) << "db: adding sent message: " << fname;
     additional_sent_tags.insert (additional_sent_tags.end (), sent_tags.begin (), sent_tags.end ());
+
+    if (!parent_mid.empty () &&
+            find(additional_sent_tags.begin(), additional_sent_tags.end(), "*") != additional_sent_tags.end()) {
+        notmuch_message_t * parent_msg;
+        notmuch_database_find_message (nm_db, parent_mid.c_str (), &parent_msg);
+        vector<ustring> parent_tags;
+        for (notmuch_tags_t * tags = notmuch_message_get_tags (parent_msg); notmuch_tags_valid (tags); notmuch_tags_move_to_next (tags)) {
+            parent_tags.push_back (notmuch_tags_get (tags));
+        }
+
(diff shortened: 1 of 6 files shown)


src/db.cc
```diff
@@ -327,14 +327,39 @@ namespace Astroid {
...
   }
 
-  ustring Db::add_sent_message (ustring fname, vector<ustring> additional_sent_tags) {
+  ustring Db::add_sent_message (ustring fname, vector<ustring> additional_sent_tags, ustring parent_mid) {
     LOG (info) << "db: adding sent message: " << fname;
     additional_sent_tags.insert (additional_sent_tags.end (), sent_tags.begin (), sent_tags.end ());
+
+    if (!parent_mid.empty () &&
+            find(additional_sent_tags.begin(), additional_sent_tags.end(), "*") != additional_sent_tags.end()) {
+        notmuch_message_t * parent_msg;
```
gauteh: Probably better to use `on_message` here, takes care of cleaning up the parent_msg. Might also be better to instantiate a `NotmuchMessage` on `parent_msg`, there is already a `get_tags` function which is more robust. `NotmuchMessage` should be cheap objects.

mxmehl:
Interesting approach! #317 goes into a similar direction, especially for initial messages starting a new thread.

Could the work of this PR make realisation of #317 easier?

gauteh:
Max writes on October 21, 2018 12:53:
> Could the work of this PR make realisation of #317 easier?

It's fairly orthogonal, but I got inspired and wrote a PR for #317: 
please test #583 for that. Post further comments on #583 on that PR.

src/db.cc
larskotthoff: Thanks, I'll have a look at that.

gauteh:
Lars Kotthoff writes on October 20, 2018 21:56:
> This PR introduces special values to `accounts.<account>.additional_sent_tags`, namely '*' (apply all tags that the message being replied to, if any, has), and '-<tag>' (to exclude tags from being applied). For example, the option "foo,*,-attachment" applies the tag "foo" and any tags the message being replied to has, except for "attachment".

Does the syntax also work for the general `mail.sent_tags`?

larskotthoff:
Good idea, I'll add that.

larskotthoff:
I've changed the code to use `NotmuchMessage`. Turns out that this syntax for the general `mail.sent_tags` was already supported because I'm doing the processing after merging in those tags.

larskotthoff:
Thanks, done!

PR #582, opened 2018-10-21
Move/edit entire tags

larskotthoff:
Enables to move over entire tags in the tag editing box with the left and right cursor keys, and delete an entire tag at a time with delete/backspace.

### DIFF
--- a/src/command_bar.cc
+++ b/src/command_bar.cc
@@ -277,6 +277,9 @@ namespace Astroid {
             ustring_sz end = entry.get_position ();
             // normal behavior if we're not in between tags
             if (txt[end] != ' ' && end > 0 && end < txt.size ()) return false;
+            // normal behavior if we're at the end and the character is not a
+            // space -- assume editing tag
+            if (txt[end] != ' ' && end == txt.size ()) return false;
             ustring_sz start = end;
             // skip two spaces at the end
             if (start == txt.size ()) start--;
--- a/src/command_bar.cc
+++ b/src/command_bar.cc
@@ -258,6 +258,8 @@ namespace Astroid {
           if (mode == CommandMode::Tag || mode == CommandMode::DiffTag) {
             ustring txt = entry.get_text ();
             ustring_sz pos = entry.get_position ();
+            // skip two spaces at the end
+            if (pos == txt.size ()) pos--;
             if (pos > 0) pos--;
             while (txt[pos] != ' ' && pos > 0) pos--;
             entry.set_position (pos);
@@ -272,6 +274,8 @@ namespace Astroid {
             ustring txt = entry.get_text ();
             ustring_sz end = entry.get_position ();
             ustring_sz start = end;
+            // skip two spaces at the end
+            if (start == txt.size ()) start--;
             if (start > 0) start--;
             while (txt[start] != ' ' && start > 0) start--;
             entry.set_text (txt.substr (0, start) + txt.substr (end, txt.size () - end));
@@ -287,7 +291,7 @@ namespace Astroid {
             ustring txt = entry.get_text ();
             ustring_sz pos = entry.get_position ();
             if (pos < txt.size ()) pos++;
-            while (txt[pos] != ' ' && pos < txt.size()) pos++;
+            while (txt[pos] != ' ' && pos < txt.size ()) pos++;
             entry.set_position (pos);
             return true;
           }
--- a/src/command_bar.cc
+++ b/src/command_bar.cc
@@ -258,6 +258,8 @@ namespace Astroid {
           if (mode == CommandMode::Tag || mode == CommandMode::DiffTag) {
             ustring txt = entry.get_text ();
             ustring_sz pos = entry.get_position ();
+            // normal behavior if we're not in between tags
+            if (txt[pos] != ' ' && pos > 0 && pos < txt.size ()) return false;
             // skip two spaces at the end
             if (pos == txt.size ()) pos--;
             if (pos > 0) pos--;
@@ -273,6 +275,8 @@ namespace Astroid {
           if (mode == CommandMode::Tag || mode == CommandMode::DiffTag) {
             ustring txt = entry.get_text ();
             ustring_sz end = entry.get_position ();
+            // normal behavior if we're not in between tags
+            if (txt[end] != ' ' && end > 0 && end < txt.size ()) return false;
             ustring_sz start = end;
             // skip two spaces at the end
             if (start == txt.size ()) start--;
@@ -290,6 +294,8 @@ namespace Astroid {
           if (mode == CommandMode::Tag || mode == CommandMode::DiffTag) {
             ustring txt = entry.get_text ();
             ustring_sz pos = entry.get_position ();
+            // normal behavior if we're not in between tags
+            if (txt[pos] != ' ' && pos > 0 && pos < txt.size ()) return false;
             if (pos < txt.size ()) pos++;
             while (txt[pos] != ' ' && pos < txt.size ()) pos++;
             entry.set_position (pos);
@@ -303,6 +309,8 @@ namespace Astroid {
           if (mode == CommandMode::Tag || mode == CommandMode::DiffTag) {
             ustring txt = entry.get_text ();
             ustring_sz start = entry.get_position ();
+            // normal behavior if we're not in between tags
+            if (txt[start] != ' ' && start > 0 && start < txt.size ()) return false;
             ustring_sz end = start;
             if (end < txt.size ()) end++;
             while (txt[end] != ' ' && end < txt.size ()) end++;
@@ -353,6 +361,9 @@ namespace Astroid {
               }
             }
 
+            return true;
+          }
+          if (mode == CommandMode::Tag || mode == CommandMode::DiffTag) {
             return true;
           }
         }
--- a/src/command_bar.cc
+++ b/src/command_bar.cc
@@ -135,52 +135,12 @@ namespace Astroid {
 
     reset_bar ();
 
-    setup_bar_labels ();
-
-    switch (mode) {
-
-      case CommandMode::AttachMids:
-        {
-          start_generic (cmd);
-        }
-        break;
-
-      case CommandMode::Search:
-        {
-          start_searching (cmd);
-        }
-        break;
-
-      case CommandMode::Filter:
-      case CommandMode::SearchText:
-        {
-          start_text_searching (cmd);
-        }
-        break;
-
-      case CommandMode::Tag:
-        {
-          start_tagging (cmd);
-        }
-        break;
-
-      case CommandMode::DiffTag:
-        {
-          start_difftagging (cmd);
-        }
-        break;
-    }
-    callback = f;
-    entry.set_position (-1);
-    set_search_mode (true);
-  }
-
-  void CommandBar::setup_bar_labels () {
     switch (mode) {
 
       case CommandMode::AttachMids:
         {
           entry.set_icon_from_icon_name ("mail-attachment-symbolic");
+          start_generic (cmd);
         }
         break;
 
@@ -188,6 +148,7 @@ namespace Astroid {
         {
           mode_label.set_text ("");
           entry.set_icon_from_icon_name ("edit-find-symbolic");
+          start_searching (cmd);
         }
         break;
 
@@ -196,31 +157,29 @@ namespace Astroid {
         {
           mode_label.set_text ("Find text");
           entry.set_icon_from_icon_name ("edit-find-symbolic");
+          start_text_searching (cmd);
         }
         break;
 
       case CommandMode::Tag:
         {
-          if (edit_mode == EditMode::Tags) {
-            mode_label.set_text ("Change tags [move by tag]");
-          } else {
-            mode_label.set_text ("Change tags");
-          }
+          mode_label.set_text ("Change tags");
           entry.set_icon_from_icon_name ("system-run-symbolic");
+          start_tagging (cmd);
         }
         break;
 
       case CommandMode::DiffTag:
         {
-          if (edit_mode == EditMode::Tags) {
-            mode_label.set_text ("Change tags (+/-) [move by tag]");
-          } else {
-            mode_label.set_text ("Change tags (+/-)");
-          }
+          mode_label.set_text ("Change tags (+/-)");
           entry.set_icon_from_icon_name ("system-run-symbolic");
+          start_difftagging (cmd);
         }
         break;
     }
+    callback = f;
+    entry.set_position (-1);
+    set_search_mode (true);
   }
 
   void CommandBar::reset_bar () {
@@ -244,7 +203,7 @@ namespace Astroid {
     entry.set_completion (search_completion);
     current_completion = search_completion;
 
-    search_completion->color_tags ();
+    search_completion->color_tags (edit_mode);
   }
 
   void CommandBar::start_text_searching (ustring searchstring) {
@@ -266,7 +225,7 @@ namespace Astroid {
 
     entry.set_text (tagstring);
 
-    tag_completion->color_tags ();
+    tag_completion->color_tags (edit_mode);
   }
 
   void CommandBar::start_difftagging (ustring tagstring) {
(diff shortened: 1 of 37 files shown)


src/command_bar.cc
```diff
@@ -38,6 +38,14 @@ namespace Astroid {
         sigc::mem_fun (this, &CommandBar::entry_key_press)
         );
 
+    entry.signal_event ().connect ([&] (GdkEvent* event)->bool {
```
gauteh: Isn't this already covered by the signal connection above?

larskotthoff: No, at least not on my machine (and it took me a while to figure this out). This event only fires when the text box loses and regains focus (happens for example when pressing the alt key, and I think this is the reason it works for everything else). I wasn't able to find anything in the documentation on this.

larskotthoff:
When I want to edit a tag I usually just delete it and then autocomplete to the right one (I find this faster). I guess I could add a configuration option for this.

gauteh:
Lars Kotthoff writes on October 21, 2018 19:06:
> When I want to edit a tag I usually just delete it and then autocomplete to the right one (I find this faster). I guess I could add a configuration option for this.

Yes, I agree. I usually do the same. But I think it would be frustrating 
for users when trying to fix simple typos though. Sometime tags are used 
in a hierarchy, e.g.: `notmuch::bugs::lib` and they can become quite 
long. In which case if you just made a typo in the middle you don't 
necessarily want to re-do it. Especially if it is the first time you 
create this tag.

So the things I think need to be solved or ruled out are:

  * When moving cursor to middle of word by using mouse, hitting 
    backspace deletes to beginning of word. Should have normal backspace 
    behavior?

  * Cannot select using Shift-Left (not important)

  * The reason the key-press signal does not work is probably that this 
    is covered by the move-cursor signal [0]. The same is probably true 
    for backspace [1].

  * Not possible to move one char. Maybe Ctrl+left could do this?

  * I noticed that Down transfers focus from the Entry, we should probably 
    map this and return true so that we stop the event.

[0] https://developer.gnome.org/gtk3/stable/GtkEntry.html#GtkEntry-move-cursor
[1] https://developer.gnome.org/gtk3/stable/GtkEntry.html#GtkEntry-backspace

larskotthoff:
Hmm, what do you think about adding a configuration option to switch between the two behaviours? I'm hesitant to introduce modifiers as this won't work for everybody -- for example I use Ctrl-<cursor keys> to switch workspaces.

larskotthoff:
Hmm, what do you think about adding a configuration option to switch between the two behaviours? I'm hesitant to introduce modifiers as this won't work for everybody -- for example I use Ctrl-<cursor keys> to switch workspaces.

larskotthoff:
Hmm, what do you think about adding a configuration option to switch between the two behaviours? I'm hesitant to introduce modifiers as this won't work for everybody -- for example I use Ctrl-<cursor keys> to switch workspaces.

gauteh:
Lars Kotthoff writes on October 22, 2018 18:02:
> Hmm, what do you think about adding a configuration option to switch between the two behaviours? I'm hesitant to introduce modifiers as this won't work for everybody -- for example I use Ctrl-<cursor keys> to switch workspaces.

Hm, ideally both should be available at the same time. One option is to 
make the keybindings configurable by setting up a Keybindings instance 
and handle the keys through it first (then fallback to entry). I don't 
think we have enough keys to get by without modifiers (without modes), 
but I agree that they should at least be configurable so that users can 
adapt to their setup.

larskotthoff:
Ok, I'll think about this (and sorry for the duplicated comments; apparently github was having some issues).

gauteh:
Lars Kotthoff writes on October 22, 2018 21:55:
> Ok, I'll think about this (and sorry for the duplicated comments; apparently github was having some issues).

Ok. Yeah, I noticed on several issues :) No problem. A minor improvement 
I was thinking about, but which can wait for later: when starting 
to tag there's an extra space so that it is easy to just keep writing. 
Would be nice if Left still worked and moved to the beginning of the 
previous tag regardless of trailing whitespace.

larskotthoff:
Ok, fixed the down arrow issue and the double space at the end of the tags.

I've changed the code to allow normal movement and editing when not on a tag boundary, i.e. you click inside a tag and you can delete individual characters, you're in between tags (or at the start or end) and movement and deletion works per tag. What do you think?

gauteh:
Lars Kotthoff writes on October 23, 2018 4:51:
> Ok, fixed the down arrow issue and the double space at the end of the tags.

Nice.

> I've changed the code to allow normal movement and editing when not on a tag boundary, i.e. you click inside a tag and you can delete individual characters, you're in between tags (or at the start or end) and movement and deletion works per tag. What do you think?

Ok! That's good. I think that without a way of doing it completely with 
the keyboard we need a config option like you suggested earlier.

larskotthoff:
Minor tweak: when pressing backspace at the end of the text and not on a backspace, normal behavior. This is useful when entering a tag to be able to fix a one-character typo or similar.

What are your thoughts on the config option? I'm slowly convincing myself that this would be a bad idea as this is really something you'd want to be able to switch on the fly.

larskotthoff:
Thanks, I'll think about this. I think something that keeps it as simple as possible is probably best so the user doesn't get confused.

How about having the standard behavior (i.e. move over characters) by default and only if the user hits a modifier (e.g. control) switch to the move-over-tags behavior. This could be reset automatically when the user starts typing. We'd need some way to show the user what mode we're in though.

gauteh:
Lars Kotthoff writes on October 28, 2018 18:30:
> Thanks, I'll think about this. I think something that keeps it as simple as possible is probably best so the user doesn't get confused.
> 
> How about having the standard behavior (i.e. move over characters) by default and only if the user hits a modifier (e.g. control) switch to the move-over-tags behavior. This could be reset automatically when the user starts typing. We'd need some way to show the user what mode we're in though.

Yeah, I'd probably use pretty much only the move-over-tags behavior. 
Simple is good. A visual cue about mode is a good idea. Could change 
icon, background color.. add a border around the tags? Or use italic 
font?

larskotthoff:
I've implemented two modes -- one with the old behavior (move/edit by character), and one with the new. Switch between the two by pressing Ctrl (left or right), new option `general.tagbar_move` that controls the default behavior. If it's "tag" new behavior, else old. If unset, default to new behavior.

The mode is indicated in the label (adds `[move by tag]` if in new behavior mode). I've had a quick look and it wasn't obvious to me how to add borders around the tags (and the appropriate color would have to be determined somehow based on the color of the tag), so I didn't add that yet. Could be added later though.

gauteh:
Lars Kotthoff writes on October 30, 2018 4:15:
> I've implemented two modes -- one with the old behavior (move/edit by character), and one with the new. Switch between the two by pressing Ctrl (left or right), new option `general.tagbar_move` that controls the default behavior. If it's "tag" new behavior, else old. If unset, default to new behavior.
> 
> The mode is indicated in the label (adds `[move by tag]` if in new behavior mode). I've had a quick look and it wasn't obvious to me how to add borders around the tags (and the appropriate color would have to be determined somehow based on the color of the tag), so I didn't add that yet. Could be added later though.

Great! Works well as far as I can see. This will be super-useful! I am 
tempted to merge right away, but I think we should change the [move by 
tag] indicator to something that does not take so much space and which 
doesn't resize the entry when toggled. Almost better with no indicator! 
What do you think? I know that I am being pedantic, but it is a very 
important GUI element.

Ideally it should be possible for the user to somehow know that the mode 
can be changed with Ctrl.. but I don't have any good ideas for how.

larskotthoff:
Thanks for your comments -- it's great that you're thinking about this in so much detail; I agree that this is important.

I think having a border around the tags when move-by-tag is enabled would be much better than the text indicator -- I'll have a go at implementing that.

How about including the Ctrl shortcut in the help page? This is where users would get information on what the shortcut for editing tags is, so it seems like a natural place to have it. So then the question is whether we should allow the user to change this, but as far as I can see nothing but Ctrl really makes sense for this.

gauteh:
Lars Kotthoff writes on November 2, 2018 19:10:
> Thanks for your comments -- it's great that you're thinking about this in so much detail; I agree that this is important.
> 
> I think having a border around the tags when move-by-tag is enabled would be much better than the text indicator -- I'll have a go at implementing that.

If that's impossible then maybe swapping between italic font and not.

> How about including the Ctrl shortcut in the help page? This is where users would get information on what the shortcut for editing tags is, so it seems like a natural place to have it. So then the question is whether we should allow the user to change this, but as far as I can see nothing but Ctrl really makes sense for this.

Yes, can be included in the search keybinding. Also, could add it as a 
tooltip to the entry. Then users would see it if they try to use the 
mouse to get into the tag.

gauteh:
I think these are the available attributes: https://developer.gnome.org/pangomm/unstable/classPango_1_1Attribute.html

I made a short try at using the underline property to highlight the currently active tag, but doesn't really work, posting here for reference:

```
diff --git a/src/command_bar.cc b/src/command_bar.cc
index 8f494ec7..a8356adb 100644
--- a/src/command_bar.cc
+++ b/src/command_bar.cc
@@ -244,7 +244,7 @@ namespace Astroid {
     entry.set_completion (search_completion);
     current_completion = search_completion;
 
-    search_completion->color_tags ();
+    search_completion->color_tags (edit_mode);
   }
 
   void CommandBar::start_text_searching (ustring searchstring) {
@@ -266,7 +266,7 @@ namespace Astroid {
 
     entry.set_text (tagstring);
 
-    tag_completion->color_tags ();
+    tag_completion->color_tags (edit_mode);
   }
 
   void CommandBar::start_difftagging (ustring tagstring) {
@@ -277,7 +277,7 @@ namespace Astroid {
 
     entry.set_text (tagstring);
 
-    difftag_completion->color_tags ();
+    difftag_completion->color_tags (edit_mode);
   }
 
   bool CommandBar::entry_key_press (GdkEventKey * event) {
@@ -450,7 +450,7 @@ namespace Astroid {
     } else if (mode == CommandMode::Tag || mode == CommandMode::DiffTag || mode == CommandMode::Search) {
 
       if (current_completion)
-        refptr<TagCompletion>::cast_dynamic (current_completion)->color_tags ();
+        refptr<TagCompletion>::cast_dynamic (current_completion)->color_tags (edit_mode);
 
     }
   }
@@ -640,7 +640,7 @@ namespace Astroid {
     return true;
   }
 
-  void CommandBar::TagCompletion::color_tags () {
+  void CommandBar::TagCompletion::color_tags (EditMode edit_mode) {
     Gtk::Entry * entry = get_entry ();
     if (entry == NULL) return;
 
@@ -676,7 +676,7 @@ namespace Astroid {
       /* grapheme positions */
       ustring_sz gstart = txt.substr (0, pos).bytes ();
 
-      color_tag (tag, gstart, attrs);
+      color_tag (tag, gstart, attrs, entry->get_position(), edit_mode);
 
       pos = end+1;
     }
@@ -684,9 +684,13 @@ namespace Astroid {
     entry->set_attributes (attrs);
   }
 
-  void CommandBar::TagCompletion::color_tag (ustring tag,
-    ustring_sz gstart, Pango::AttrList &attrs) {
-
+  void CommandBar::TagCompletion::color_tag (
+      ustring tag,
+      ustring_sz gstart,
+      Pango::AttrList &attrs,
+      ustring_sz cursor,
+      EditMode _edit_mode)
+  {
     unsigned char cv[3] = { (unsigned char) (canvas_color.get_red_u ()   * 255 / 65535),
                             (unsigned char) (canvas_color.get_green_u () * 255 / 65535),
                             (unsigned char) (canvas_color.get_blue_u ()  * 255 / 65535) };
@@ -710,6 +714,16 @@ namespace Astroid {
     bgalpha.set_start_index (gstart);
     bgalpha.set_end_index   (gend);
 
+    if (_edit_mode == EditMode::Tags &&
+        cursor >= gstart &&
+        cursor+1 <= gend)
+    {
+      auto underline = Pango::Attribute::create_attr_underline (Pango::UNDERLINE_DOUBLE);
+      underline.set_start_index (gstart);
+      underline.set_end_index (gend);
+      attrs.insert (underline);
+    }
+
     attrs.insert (bga);
     attrs.insert (bgalpha);
     attrs.insert (fga);
@@ -845,7 +859,7 @@ namespace Astroid {
     return true;
   }
 
-  void CommandBar::SearchCompletion::color_tags () {
+  void CommandBar::SearchCompletion::color_tags (EditMode edit_mode) {
     Gtk::Entry * entry = get_entry ();
     if (entry == NULL) return;
 
@@ -885,7 +899,7 @@ namespace Astroid {
       /* grapheme positions */
       ustring_sz gstart = txt.substr (0, pos).bytes ();
 
-      color_tag (tag, gstart, attrs);
+      color_tag (tag, gstart, attrs, entry->get_position(), edit_mode);
 
       pos = end+1;
     }
diff --git a/src/command_bar.hh b/src/command_bar.hh
index dc249e22..aa2b2e49 100644
--- a/src/command_bar.hh
+++ b/src/command_bar.hh
@@ -117,8 +117,13 @@ namespace Astroid {
           static Gdk::RGBA  canvas_color;
           static bool       canvas_color_set;
 
-          virtual void color_tags ();
-          void color_tag (ustring tg, ustring_sz start, Pango::AttrList &attrs);
+          virtual void color_tags (EditMode);
+          void color_tag (
+              ustring tg,
+              ustring_sz start,
+              Pango::AttrList &attrs,
+              ustring_sz cursor,
+              EditMode _em);
       };
 
       refptr<TagCompletion> tag_completion;
@@ -160,7 +165,7 @@ namespace Astroid {
 
           bool on_match_selected(const Gtk::TreeModel::iterator& iter) override;
 
-          void color_tags () override;
+          void color_tags (EditMode) override;
       };
 
       refptr<SearchCompletion> search_completion;
```

larskotthoff:
Thanks, I had a look at it but it seems that adding boxes would at the very least make the code much more complex.

I'm starting to think that this may require changing the way tags are edited completely, i.e. instead of a text box have a container of text boxes for each tag individually.

gauteh:
Lars Kotthoff writes on November 4, 2018 20:15:
> Thanks, I had a look at it but it seems that adding boxes would at the very least make the code much more complex.

Yes, maybe underlinging which is already available is a good 
alternative.


> I'm starting to think that this may require changing the way tags are edited completely, i.e. instead of a text box have a container of text boxes for each tag individually.

I think this will be pretty complex, and probably unlikely that it will 
look very good. You would probably have to draw on a canvas manually. 
But then you'd have to re-implement all the text-input logic. I'm not 
completely against it, but I think it would be better to solve it within 
the capabilities of the Gtk::Entry.

larskotthoff:
Implemented the underlining. Please have a look.

gauteh:
Lars Kotthoff writes on November 6, 2018 3:31:
> Implemented the underlining. Please have a look.

Thanks! I changed the underlining to be shown when editing in char mode 
(hope you don't mind), and merged.

Please add update the wiki with the new config option.

larskotthoff:
My intention for the underlining was for it to show that the tag is one unit, to indicate that it will treated as such when moving etc. I find it counter-intuitive to show it in char mode.

gauteh:
Lars Kotthoff writes on November 6, 2018 19:30:
> My intention for the underlining was for it to show that the tag is 
> one unit, to indicate that it will treated as such when moving etc. I 
> find it counter-intuitive to show it in char mode.

Yes, that's a good point. I would like to keep the default 
tag-editing as clean as possible. An alternative is to have only the 
current tag underlined, indicating that the entire tag will be skipped 
or deleted.

larskotthoff:
I was thinking about that as well, but as moving happens over entire tags, there isn't really a current tag (cursor will be in between tags). Another option would be to underline the tag before and after, but that's much more complex to implement than underlining all tags.

Maybe an alternative would be to make the old behavior the default, with no underlining?

larskotthoff:
Well, Del will delete the next tag, so in a sense it's also the next tag :)

I really think that using the underlining to denote that a tag will be treated as a unit is much more intuitive than the other way round.

gauteh:
Somehow backspace doesn't work for me anymore, does it still for you?

PR #588, opened 2018-10-28
fix #587: only messages that are being read should be processed

### LINKED ISSUE #587: process message: process message plugin should not always be applied
The process message plugin (introduced through #571, #577, 501384f6f1719dbbf94f2dfdf1a67bd0006ba2de)  should not be applied to `ComposeMessage::build`. 

(Cc: @larskotthoff )

gauteh:
fixes #587.

### DIFF
--- a/src/compose_message.cc
+++ b/src/compose_message.cc
@@ -261,7 +261,7 @@ namespace Astroid {
 
   void ComposeMessage::load_message (ustring _mid, ustring fname) {
     set_id (_mid);
-    Message msg (_mid, fname);
+    UnprocessedMessage msg (_mid, fname);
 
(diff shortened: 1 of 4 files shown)


gauteh:
@larskotthoff can you take a look if this fixes some of the issues with the `process` plugin function?

larskotthoff:
Thanks a lot for looking at this -- I've been meaning to get round to it. As far as I can tell it works great and fixes the issues.

gauteh:
Lars Kotthoff writes on October 28, 2018 18:47:
> Thanks a lot for looking at this -- I've been meaning to get round to it. As far as I can tell it works great and fixes the issues.

Great, thanks for looking over. Merging.

- gaute

PR #589, opened 2018-10-29
Trash threads in index and view modes

e11o:
Just added the ability to trash threads (toggle 'trash' tag) in index and view modes.

Added potential keybindings following the pattern for archiving threads (which I consider to be the closest action in nature) but left most unbound because it looked a bit overkill and didn't want to take over all remaining keys. In the only few places where I bound a key, I used GDK_KEY_Delete (as "d" and "t" were already taken).

### DIFF
--- a/src/modes/thread_view/thread_view.cc
+++ b/src/modes/thread_view/thread_view.cc
@@ -1385,6 +1385,20 @@ namespace Astroid {
           return true;
         });
 
+    keys.register_key (Key (GDK_KEY_BackSpace),
+        "thread_view.trash_thread",
+        "Toggle 'trash' tag on the whole thread",
+        [&] (Key) {
+
+          if (!edit_mode && focused_message) {
+            if (thread) {
+              main_window->actions->doit (refptr<Action>(new TrashAction(thread)));
+            }
+          }
+
+          return true;
+        });
+
     keys.register_key ("C-P",
         "thread_view.print",
         "Print focused message",
@@ -1568,6 +1582,16 @@ namespace Astroid {
           return true;
         });
 
+    next_multi.register_key (UnboundKey(),
+        "thread_view.multi_next_thread.trash",
+        "Trash and goto next",
+        [&] (Key) {
+          keys.handle ("thread_view.trash_thread");
+          emit_index_action (IA_Next);
+
+          return true;
+        });
+
     next_multi.register_key (Key ("A"),
         "thread_view.multi_next_thread.archive_next_unread_thread",
         "Archive, goto next unread",
@@ -1578,11 +1602,31 @@ namespace Astroid {
           return true;
         });
 
+    next_multi.register_key (UnboundKey(),
+        "thread_view.multi_next_thread.trash_next_unread_thread",
+        "Trash and goto next unread",
+        [&] (Key) {
+          keys.handle ("thread_view.archive_trash");
+          emit_index_action (IA_NextUnread);
+
+          return true;
+        });
+
     next_multi.register_key (Key ("x"),
         "thread_view.multi_next_thread.close",
         "Archive, close",
         [&] (Key) {
           keys.handle ("thread_view.archive_thread");
+          close ();
+
+          return true;
+        });
+
+    next_multi.register_key (Key (GDK_KEY_Delete),
+        "thread_view.multi_next_thread.trash_close",
+        "Trash and close",
+        [&] (Key) {
+          keys.handle ("thread_view.trash_thread");
           close ();
 
           return true;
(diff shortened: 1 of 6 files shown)


mxmehl:
Since you asked for others' experiences: I personally also use the tag `del` to move things to the trash bin after the next run of poll.sh and a [custom script](https://src.mehl.mx/mxmehl/mail-config/src/branch/master/astroid/scripts/move-tag.sh#L59) which takes care of moving mails based on their tags.

The same applies to the spam tag which I also use frequently, and for which there is already a built-in keybinding (which I [changed](https://src.mehl.mx/mxmehl/mail-config/src/branch/master/astroid/keybindings#L19)).

Obviously, the `del` and `spam` tags are excluded tags within notmuch.

For actually deleting single emails within threads, for example vacation auto-replies, I use [a hook](https://src.mehl.mx/mxmehl/mail-config/src/branch/master/astroid/hooks/delmail).

I'm not using the archive tag at all so I can't say anything about its usage in astroid or notmuch.

e11o:
To me, it's almost the same concept as archiving (removing the "inbox" tag). It would work out of the box with some servers (like Gmail) but will need some wrangling with others (Office365, anyone?). For the second case, it would at least mark the threads so something else (like alot) can pick them up.

I don't think all functionality can be implemented right now via hooks. For example, delete and close in thread_view mode and toggling ("t") multiple threads in thread_index mode and deleting them together.

If we want to be on the safer side, we can leave all actions with unbound keys and document for those who want the feature. Probably no one would use Delete or BackSpace for this anyway ("d" is much more appealing) so they will need to config keybindings anyway.

src/modes/thread_view/thread_view.cc
```diff
@@ -1385,6 +1385,20 @@ namespace Astroid {
           return true;
         });
 
+    keys.register_key (Key (GDK_KEY_BackSpace),
```
gauteh: Why is it sometimes Backspace and sometimes Delete? Should always be the same I think. I think that we can use `d`

varac:
@e11o What's the state of the PR ? @gauteh proposed changes to your PR, can you please consider them so we can merge this ?

e11o:
Sorry that I totally dropped the ball here but I will complete the PR now. There's a case that doesn't work as expected though. 

Previously trashed threads will appear again in the thread index when a new email is added to the thread. This is desired behavior.

But I would expect thrashing would just trash the thread again but, as it's implemented using a ToggleAction, it will actually delete the existing trash tag. You actually need to trash twice to delete the thread.

On the other hand, there might be a valid use case for removing the trash tag in thread index mode. As trash would typically be added to search.exclude_tags in notmuch, this use case would be when searching explicitly with tag:trash. Restoring emails is probably the most common scenario.

So it seems like the action in thread_index needs to be contextual and never a toggle. If tag:trash is present in the search string then it should remove the trash tag. If tag:trash is not present then it should remove and add the tag again?

@gauteh Does that make sense? Is it possible to know tags in the search string when executing an action?

src/modes/thread_view/thread_view.cc
e11o: I will look again but I believe `d` is already taken in the default keybindings for one of the modes.

gauteh:
Would it make sense to:

if all messages has trash: untrash
if some, but not all, or none has trash: trash

this seems to give the same behavior, and I think would be more deterministic.

e11o:
I like it better, yes.

Alternatively, could use a binding for sending to trash and another for restoring. Both would be no-op in case there's nothing to trash or restore.

Would be more explicit but requires 2 bindings...

e11o:
@gauteh I updated the key bindings to use "d". Had to also use Delete for one particular case.

Regarding the `if all messages has trash: untrash` approach, I originally thought you were refering to "all selected messages" but that wouldn't resolve the problem.
Now I believe you were referring to check all threads in the list store. Wouldn't that be very inefficient for a large result set?

varac:
@gauteh  Lets get this merged soon please, waiting for this feature ! Thanks !

gauteh:
Esteban Sancho writes on April 18, 2020 23:16:
> @gauteh I updated the key bindings to use "d". Had to also use Delete
> for one particular case.
>
> Regarding the `if all messages has trash: untrash` approach, I
> originally thought you were refering to "all selected messages" but
> that wouldn't resolve the problem.
> Now I believe you were referring to check all threads in the list
> store. Wouldn't that be very inefficient for a large result set?

I meant to check all the _messages_ of a thread, not _threads_ in the
thread-index. So if a _thread_ is trashed, but has a new _untrashed_
message appended this will cause the new message to get `trash` tag
added -> causing the entire thread to be trashed. Otherwise, if all
_messages_ have `trash` tag: untrash thread (i.e. remove trash from all
messages). Notmuch does not store tags on threads, but on messages. That
way you are not dependent on the particular query you are in.

PR #593, opened 2018-11-07
Cleanup webkit references

larskotthoff:
The destructor for the thread view object didn't release all references, which lead to webkit render processes not being cleaned up after closing the thread view.

### DIFF
--- a/src/modes/edit_message.cc
+++ b/src/modes/edit_message.cc
@@ -545,8 +545,6 @@ namespace Astroid {
     if (is_regular_file (tmpfile_path)) {
       boost::filesystem::remove (tmpfile_path);
     }
-
-    delete thread_view;
   }
 
   void EditMessage::pre_close () {
--- a/src/modes/edit_message.cc
+++ b/src/modes/edit_message.cc
@@ -212,7 +212,7 @@ namespace Astroid {
     }
 # endif
 
-    thread_view = new ThreadView(main_window, true);
+    thread_view = Gtk::manage(new ThreadView(main_window, true));
     //thread_rev->add (*thread_view);
     editor_box->pack_start (*thread_view, false, false, 2);
 
--- a/src/modes/edit_message.cc
+++ b/src/modes/edit_message.cc
@@ -212,7 +212,7 @@ namespace Astroid {
     }
 # endif
 
-    thread_view = Gtk::manage(new ThreadView(main_window, true));
+    thread_view = new ThreadView(main_window, true);
     //thread_rev->add (*thread_view);
     editor_box->pack_start (*thread_view, false, false, 2);
 
@@ -545,6 +545,12 @@ namespace Astroid {
     if (is_regular_file (tmpfile_path)) {
       boost::filesystem::remove (tmpfile_path);
     }
+
+    delete thread_view;
+  }
+
+  void EditMessage::pre_close () {
+    ((Mode*) thread_view)->pre_close ();
   }
 
   void EditMessage::close (bool force) {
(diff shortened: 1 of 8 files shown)


larskotthoff:
It's possible that this should be done elsewhere, but as far as I can tell this isn't cleaned up otherwise -- I'm getting a stale `WebKitNetworkProcess` for each closed thread view that is only terminated when astroid quits otherwise.

I've also just noticed that the same thing happens with the render process after sending a message, but haven't been able to track down where the refs should be freed yet.

gauteh:
Lars Kotthoff writes on November 7, 2018 19:38:
> It's possible that this should be done elsewhere, but as far as I can tell this isn't cleaned up otherwise -- I'm getting a stale `WebKitNetworkProcess` for each closed thread view that is only terminated when astroid quits otherwise.
> 
> I've also just noticed that the same thing happens with the render process after sending a message, but haven't been able to track down where the refs should be freed yet.

The WebView is a initially unowned object so it should not be unreffed 
unless we sink it, the other two can be unreffed after they have been 
reffed by the webview. Since we keep the webview ref around we should 
sink it and unref it in destructor.

See: 

http://www.cs.hunter.cuny.edu/~sweiss/course_materials/csci493.70/lecture_notes/GTK_memory_mngmt.pdf

   and 

https://webkitgtk.org/reference/webkit2gtk/stable/WebKitSettings.html

Something like:
```

diff --git a/src/modes/thread_view/thread_view.cc b/src/modes/thread_view/thread_view.cc
index 9843cde4..19b4eb8a 100644
--- a/src/modes/thread_view/thread_view.cc
+++ b/src/modes/thread_view/thread_view.cc
@@ -112,8 +112,13 @@ namespace Astroid {
           "settings", websettings,
           NULL));
 
+    g_object_ref_sink (webview);
+    g_object_unref (context);
+    g_object_unref (websettings);
+
     gtk_box_pack_start (GTK_BOX (this->gobj ()), GTK_WIDGET (webview), true, true, 0);
 
+
     g_signal_connect (webview, "load-changed",
         G_CALLBACK(ThreadView_on_load_changed),
         (gpointer) this );
@@ -141,7 +146,7 @@ namespace Astroid {
 
   ThreadView::~ThreadView () { //
     LOG (debug) << "tv: deconstruct.";
-    g_object_unref (websettings);
+    g_object_unref (webview);
   }
 
   void ThreadView::pre_close () {
```

src/modes/edit_message.cc
```diff
@@ -212,7 +212,7 @@ namespace Astroid {
     }
 # endif
 
-    thread_view = Gtk::manage(new ThreadView(main_window, true));
```
gauteh: `ThreadView` is sub-classed from a `Gtk::Object`

src/modes/thread_index/thread_index.cc
```diff
@@ -32,10 +32,10 @@ namespace Astroid {
     list_store = Glib::RefPtr<ThreadIndexListStore>(new ThreadIndexListStore ());
     queryloader.list_store = list_store;
 
-    list_view  = Gtk::manage(new ThreadIndexListView (this, list_store));
```
gauteh: Same here. All these are subclassed from `Mode` which is a `Gtk::Box` or similar.

larskotthoff: But none of these are added to another GTK object, or did I miss something?

gauteh: It's added to the scroll window, which is added through the add_pane call. Might still be cleaned up though.. But all Gtk Widgets should either be used or it should be removed.

larskotthoff: Ok, should be reverted now.

gauteh:
Thanks, merged.

larskotthoff:
Thanks! I'm still trying to track down a memory leak that seems to occur when refreshing and new mail is found. I'll open a PR for this when I've figured it out.

gauteh:
Great, thanks! Awesome work.

PR #595, opened 2018-11-09
Refresh threadview

larskotthoff:


larskotthoff:
Expose keyboard shortcut to refresh thread view. This is useful after replying to a mail from the thread view to see the new mail.

src/modes/thread_view/thread_view.cc
```diff
@@ -554,17 +554,20 @@ namespace Astroid {
   void ThreadView::register_keys () { // {{{
     keys.title = "Thread View";
 
-# ifdef DEBUG_WEBKIT
     keys.register_key ("C-r", "thread_view.reload",
```
gauteh: I think we should use the `$` (or at least alias it as well) key since that means reload in thread-index. It is possible to handle add message or update messages events, but haven't gotten that far yet. That would maybe give a smoother experience. Have to insert the message at the correct spot though, and maybe give it some different fading background color.

larskotthoff: Happy to change that -- I've changed the keybinding for this anyway :)

And yes, the eventual intention is to have an option to automatically reload the thread view when the message has been sent, but I need to look into it a bit further to implement that. The new message is already inserted at the right spot; the only thing that's not optimal is that it is automatically focused instead of whatever message previously was (which will be hidden).

gauteh: Yes, take a look at how a message is updated if the tags are changed. It is triggered at each message from the MessageThread. In the code that loops over the messages and emits the `MESSAGE_TAGS_CHANGED` event you can also check for added and removed messages and update the DOM tree through either `add_message` or `remove_message` in `PageClient`. `add_message`, however, would not insert the message at the correct spot which probably means that you need to figure out a way to do this. Additionally, the state and messagethread vectors must match both in `ThreadView` and in the web extension.

gauteh:
Thanks, merged!

Issue #556, opened 2018-09-26
testsuite failures on some archs when building 0.13 (+ 7 patches) on Debian


jonassmedegaard:
Hi,

I recently cherry-picked a few patches and rebuild Astroid for Debian, and it succeeds on arm64 and some other archs, but fails on others: https://buildd.debian.org/status/package.php?p=astroidmail&suite=sid

The patches applied are these: https://salsa.debian.org/debian/astroidmail/tree/master/debian/patches

If you don't want to be bothered with issues in a codebase deviating from your work, then I perfectly understand that.  Sharing in case you (like me) consider the deviation minor and find the revealed testsuite failures helpful.

gauteh:
The tests are partly failing because of missing gpg-agent, did we 
discuss this issue already? Check the issue tracker. Seem to remember 
something about it with relation to packaging.

Does the same happen if you apply all patches (just to make sure that 
any fixes to the test-suite also were applied).

Regards, Gaute

jonassmedegaard:
Quoting Gaute Hope (2018-09-26 15:08:28)
> The tests are partly failing because of missing gpg-agent, did we 
> discuss this issue already? Check the issue tracker. Seem to remember 
> something about it with relation to packaging.

You suggested missing gpg-agent being the cause in the past, yes.

I might in our conversation have acknowledged that, but then looked 
closer and confirmed that was not the case.

Also (as I recall different from my earlier testsuite failures) it fails 
only for _some_ architectures, not all: I am not convinced that your 
testsuite skips the need for properly setup gpg-agent on amd64 and 
armel, but not on armhf ;-)


> Does the same happen if you apply all patches (just to make sure that 
> any fixes to the test-suite also were applied).

I have not packaged HEAD of development, no.  That is not easy to do, so 
I would prefer not needing to do that.


 - Jonas

-- 
 * Jonas Smedegaard - idealist & Internet-arkitekt
 * Tlf.: +45 40843136  Website: http://dr.jones.dk/

 [x] quote me freely  [ ] ask before reusing  [ ] keep private

gauteh:
Jonas Smedegaard writes on September 26, 2018 17:09:
> Quoting Gaute Hope (2018-09-26 15:08:28)
>> The tests are partly failing because of missing gpg-agent, did we 
>> discuss this issue already? Check the issue tracker. Seem to remember 
>> something about it with relation to packaging.
> 
> You suggested missing gpg-agent being the cause in the past, yes.
> 
[...]

PR #603, opened 2018-11-23
Shorten GNUPGHOME length

frediz:
On ppc64el, due to a bit too long GNUPGHOME, gpg socket file name is
too long and gpg fails to create foo1.key in tests/run_test.sh and
the script exits righ away with set -e, leaving the test directory
unproperly set up (missing ui directory, and keys) which makes some tests
fail (generic, theme, crypto)
Here is the Debian bug I opened :
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=909190
https://buildd.debian.org/status/fetch.php?pkg=astroidmail&arch=ppc64el&ver=0.14-1&stamp=1538432617&raw=0
https://buildd.debian.org/status/package.php?p=astroidmail

Some details here about too long socket and gpg :
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=847206#10

### DIFF
--- a/tests/run_test.sh
+++ b/tests/run_test.sh
@@ -12,7 +12,7 @@ echo "Source dir: ${SRCDIR}"
 echo "Build dir:  ${BINDIR}"
 
 export NOTMUCH_CONFIG="${BINDIR}/tests/mail/test_config"
-export GNUPGHOME="${BINDIR}/tests/test_home/gnupg"
+export GNUPGHOME="${BINDIR}/gnupg"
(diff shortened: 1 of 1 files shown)


gauteh:
Thanks! Nice catch, it seems that #556 is caused by this bug as well. It seems like a hacky fix, but also very hard to fix properly. Will test a bit, but I think this is the only solution we have.

frediz:
Indeed, I would have like to come up with something nicer, but didn't find much better in a reasonable amount of time for me :)
Thanks for your time.

PR #606, opened 2018-11-28
improve extraction of first names for thread index

larskotthoff:
This is mostly to fix two cases. First where the name is in the format "lastName, firstName", where "lastName" would be extracted before. Second "prefix. firstName lastName", where "prefix" would be extracted before (e.g. title or first initial).

I removed the code that handles the all-read case because notmuch returns a comma-separated list of names, which breaks things if the name itself contains a comma (e.g. "lastName, firstName").

### DIFF
--- a/src/db.cc
+++ b/src/db.cc
@@ -643,31 +643,6 @@ namespace Astroid {
      * an unread message in the thread */
     vector<tuple<ustring, bool>> aths;
 
-    /* first check if any messages are unread, if not: just fetch the authors for
-     * the thread without checking each one.
-     *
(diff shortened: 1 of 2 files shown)


gauteh:
The reason for including the notmuch call was that it presumably was faster than traversing the message tree. Is there any significant performance loss with this?

larskotthoff:
I haven't noticed anything. When I did some profiling in another context, I found that the notmuch database operations are much slower than anything else. My largest query has about 20,000 mails and even with this change it opens instantaneously.

gauteh:
Alright, thanks. Merged.

PR #607, opened 2018-11-28
Memory, performance, stability improvements

larskotthoff:
- fixed memory leak in database code -- `notmuch_database_close` closes the connection, but doesn't free the memory; `notmuch_database_destroy` does both
- improved performance by reducing database open wait time from 1 second to 100 milliseconds
- adjusted some log levels (e.g. "warning" when key binding was overwritten by user)
- fixed IPC to not report error when end of stream is reached
- misc memory improvements through explicitly closing things and unconditionally freeing memory
- fixed some code formatting
- fixed some compiler warnings

### DIFF
--- a/src/modes/thread_view/webextension/ae_protocol.cc
+++ b/src/modes/thread_view/webextension/ae_protocol.cc
@@ -139,15 +139,9 @@ namespace Astroid {
     gsize read = 0;
     bool  s    = false;
 
-    AeProtocol::MessageTypes mt;
-
     /* read message size */
     gsize msg_sz = 0;
     s = istream->read_all ((char *) &msg_sz, sizeof (msg_sz), read, reader_cancel);
-
-    if (s && read == 0) {
-      return mt;
-    }
 
     if (!s || read != sizeof (msg_sz)) {
       throw ipc_error ("could not read message size");
@@ -157,6 +151,7 @@ namespace Astroid {
       throw ipc_error ("message exceeds maximum size.");
     }
 
+    AeProtocol::MessageTypes mt;
     s = istream->read_all ((char*) &mt, sizeof (mt), read, reader_cancel);
 
     if (!s || read != sizeof (mt)) {
--- a/src/modes/thread_view/webextension/ae_protocol.cc
+++ b/src/modes/thread_view/webextension/ae_protocol.cc
@@ -139,9 +139,15 @@ namespace Astroid {
     gsize read = 0;
     bool  s    = false;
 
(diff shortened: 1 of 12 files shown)


larskotthoff:
Oh and this also fixes updating the titles for thread indices -- a '%' is added to indicate loading, but when loading finished the message to update the title was sent before setting the variable that indicated that loading was done. This worked sometimes.

gauteh:
Thanks a lot for these PR's, `_destroy` in particular should be good!! Will
take a look as soon as I get a chance.

larskotthoff:
Thanks, this is particularly important if you have lots of different tabs open -- I have 8 for different mailboxes etc and astroid would crash with an out of memory error after a few hours and about 1GB.

I've noticed that tagging seems to take a while with this many tabs open, presumably because of the syncing and messages being sent back and forth. Looking into that next.

src/modes/thread_view/webextension/ae_protocol.cc
```diff
@@ -139,10 +139,16 @@ namespace Astroid {
...
     bool  s    = false;
 
+    AeProtocol::MessageTypes mt;
+
     /* read message size */
     gsize msg_sz = 0;
     s = istream->read_all ((char *) &msg_sz, sizeof (msg_sz), read, reader_cancel);
 
+    if (s && read == 0) {
+      return mt;
```
gauteh: All in all, this looks very good. I'm not sure about this line, if the read function fails then I think that the caller should handle the exception more gracefully and just exit rather than receiving an empty MessageType. It is not clear to the calling thread whether the read call succeded or not.

larskotthoff: Yes, there seems to be more going on here. I regularly get log messages about broken pipes, so it seems that the read end is closed before all data has been pushed into the pipe on the write end and messages are lost. I plan to have a look at that.

Technically this case here isn't an exception; it covers the case where the stream has been closed. This is a perfectly valid case IMHO and shouldn't return an error.

gauteh: I agree, it is not an error. I just think that it should be handled by the reader thread which can then shut it self down, it could also be passed with a CloseStream messagetype.

larskotthoff: Ok, maybe postpone this until I've had a chance to look into why messages are dropped.

gauteh:
Ok, sounds good.

gauteh:
Great, thanks, merged!

PR #612, opened 2018-12-08
fix broken link in example script

arvindsv:
https://github.com/astroidmail/astroid/wiki/Processing-mail doesn't exist.


gauteh:
Thanks, merging.

PR #613, opened 2018-12-15
config.cc: Default astroid.notmuch_config to $NOTMUCH_CONFIG

glacambre:
By default, notmuch uses the $NOTMUCH_CONFIG environment variable in
order to find its config file. This commit makes astroid's default
config generator use this variable in order to populate the
astroid.notmuch_config setting.


gauteh:
I think this looks good. For some reason I thought this functionality was already included and it is already mentioned in the man page (although that sentence is pretty unclear).

Issue #380, opened 2017-08-14
"Attachment" mentioned in previous email in the thread


Radvendii:
This is by no means an urgent/important issue, but I figure it's good to have it mentioned somewhere

When I send an email with the word 'attach' or some similar word, it warns me if I don't have an actual attachment on the email. However, this also happens when for example I am replying to an email with "attach" in it. (And I have the body of the replied email quoted in my email)

If there's an easy way to ignore quoted text that would be nice.

gauteh:
Taeer Bar-Yam writes on august 14, 2017 17:55:
> If there's an easy way to ignore quoted text that would be nice.

I guess its not really that hard to check if the first char on that line 
is a '>'.

Radvendii:
Hahaha, I guess that's one way to do it. I don't have any time now, but I'll make a pull request next time I have some free time.

gauteh:
By the way: #267

Issue #267, opened 2017-02-02
strange missing attachment warning


emdete:
the warning to attach a file when the word "attach" was found in the mail and no file was attached is given even if that word is in a quoted sentence.

furthermore it even says "you have mentioned ...". maybe a more abstract wording would do for now like "astroid found the word attach in the mail while no file was attached" or the like.

gauteh:
Hi,

Yeah - I can see that it is a bit weird in this case. Optimally quoted
'attach' words wouldn't count, but for now I don't think I will
prioritize fixing that. 

I like the personal, maybe friendly, tone of how it is now. It just seems a bit
more cumbersome and distant with something like 'astroid found..'? Even
if it slightly more accurate in some situations. what do you think?

mxmehl:
+1 for adding exceptions of quoted occurences of the attachment words.

Regarding the text I don't have a strong opinion. Maybe a compromise could be "astroid found/noticed your message contains ..."?

emdete:
i consider this a ki against moderate algorithm problem 😉 just removing "you mention" is needed as astroid doesnt know. i would not pimp the algorithm to enhance detection because it always wont be sufficient.

PR #621, opened 2019-01-23
Made it not check quoted text for attachment words

Radvendii:
#380 #267

Not sure if this is the simplest way of doing this. I modified the logic from the place where it adds the quotes to reply messages.

### DIFF
--- a/src/modes/edit_message.cc
+++ b/src/modes/edit_message.cc
@@ -1045,10 +1045,22 @@ namespace Astroid {
     if (attachments.empty () && !attachment_words.empty ()) {
       ustring bl = body.lowercase ();
 
+      /* strip quoted lines */
+      ostringstream nonquoted;
+      stringstream sstr (bl);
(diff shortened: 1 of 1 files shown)


src/modes/edit_message.cc
```diff
@@ -1045,10 +1045,22 @@ namespace Astroid {
...
       ustring bl = body.lowercase ();
 
+      /* strip quoted lines */
+      ostringstream nonquoted;
+      stringstream sstr (bl);
+      while (sstr.good()) {
+        string line;
+        getline (sstr, line);
+
+        if (line[0] != '>')
```
gauteh: Can line be zero length?

Radvendii: Are strings not null terminated in C++?

gauteh: [] returns 0 on string length, so is good.

PR #624, opened 2019-02-26
Prevent segfault by reordering set_text and completion setup

theunknownxy:
When set_text is called, an completion in an inconsistent state is triggered.

For example when the CommandBar was used for filtering (ctrl+f) and is then
used for searching (for example "O"), entry_changed will try to cast
current_completion to TagCompletion. This happens because "mode" already
contains the new mode and "current_completion" contains the old value. Thus
the cast results in an nullptr, which in turn causes an segfault.


gauteh:
Thanks, merged.

Issue #319, opened 2017-04-13
Replying to messages without a text/plain portion


jimcheetham:
Similar to #236, I get a lot of email (from users of the Outlook client, via Exchange) that arrives with no text/plain portion at all.

If I'm lucky they are MIME-delineated, but I've seen them arrive with only a Content-Type main header as well.

Not having them render HTML by default seems sensible, but they're therefore difficult to preview and reply to.

When using alot, I used to be able to generate my own text versions (I believe by asking mailcap(5), which would do something like run the file through `lynx -dump`); can we do something similar here?

otwieracz:
Is this really closed? I am still not able to reply to HTML-only email, I end up quoting bunch of HTML nonsense.

gauteh:
No, not really fixed.

otwieracz:
Are there any plans how to deal with that issue? Is this somehow possible to reply to HTML-only email? Any known workaround? Turns out I receive quite a lot HTML-only emails and it become a showstopper…

agrudzinsky:
Is it possible to create a "reply hook" that would pipe the original message to a filter before inserting it into the editor?  Then this issue could be fixed by simply using "lynx -dump" or the like.  

My company mail server also inserts "[EXTERNAL]" in the subject of emails received from outside the company and warnings inside the email body.  Those get replicated in replies and create visual clutter.  I can use this "reply hook" to remove them as well.

gauteh:
Yes, it should be fairly easy to set up a plugin hooks in the same way as `process` is done now for reply messages as well. There were some recent issues on that, might find them if you search for `process` (which only applies to opening messages for viewing).

https://github.com/astroidmail/astroid/wiki/Plugins#astroidactivatable

gauteh:
Something that is a possibility is to have HTML converted to markdown when quoting, then replying in markdown. This depends on how good the html->markdown converter is. Would that solve the issue for you?

otwieracz:
Well, as you said it depends on the converter quality. I am not using it at all, so hard to tell. Front my experience however there is no such thing as “good html parser” :)

-- 
sg

gauteh:
True..

agrudzinsky:
Quoting Gaute Hope (2019-03-06 07:09:26)
> Something that is a possibility is to have HTML converted to
> markdown when quoting, then replying in markdown. This depends
> on how good the html->markdown converter is. Would that solve
> the issue for you?

Replying in HTML is not important. I prefer to reply in text.
[...]

PR #626, opened 2019-04-03
Quote html parts

gauteh:
Update `mail.reply.quote_processor` to whatever you want to process HTML parts into text when quoting.

#319 and others.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -40,7 +40,7 @@ set (CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake")
 if (CMAKE_BUILD_TYPE STREQUAL "Debug")
   # debug build - retrieve current project version from git
   execute_process (
-    COMMAND git describe --abbrev=8 --tags --always
+    COMMAND git -C "${CMAKE_SOURCE_DIR}" describe --abbrev=8 --tags --always
     OUTPUT_VARIABLE PROJECT_VERSION
     OUTPUT_STRIP_TRAILING_WHITESPACE
     )
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -69,7 +69,7 @@ Vagrant.configure("2") do |config|
   config.vm.provision "shell", inline: <<-SHELL
     pacman -Syu --noconfirm
 
-    pacman -S --noconfirm --needed cmake ninja git gcc notmuch-runtime glibmm gtkmm3 vte3 boost libsass libpeas ruby-ronn pkgconf webkit2gtk protobuf gobject-introspection xorg-xauth xorg-xclock cmark python-gobject ipython
+    pacman -S --noconfirm --needed cmake ninja git gcc notmuch-runtime glibmm gtkmm3 vte3 boost libsass libpeas ruby-ronn pkgconf webkit2gtk protobuf gobject-introspection xorg-xauth xorg-xclock cmark python-gobject ipython gvim
 
     cat > /etc/profile.d/astroid.sh <<EOL
       export ASTROID_DIR=/vagrant
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -43,20 +43,20 @@ Vagrant.configure("2") do |config|
   # the path on the host to the actual folder. The second argument is
   # the path on the guest to mount the folder. And the optional third
   # argument is a set of non-required options.
-  # config.vm.synced_folder "../data", "/vagrant_data"
+  # config.vm.synced_folder "~/.mail", "/vagrant/mail"
 
   # Provider-specific configuration so you can fine-tune various
   # backing providers for Vagrant. These expose provider-specific options.
   # Example for VirtualBox:
   #
-   config.vm.provider "virtualbox" do |vb|
-     # Display the VirtualBox GUI when booting the machine
-     # vb.gui = true
+  config.vm.provider "virtualbox" do |vb|
+    # Display the VirtualBox GUI when booting the machine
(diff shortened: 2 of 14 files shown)


mxmehl:
Thanks! I tried it and it failed for 3 HTML only mails I tested :(

Nothing happens in the UI after pressing `r`, but in the logfile there is quite some activity.

[Logfile](https://p.mehl.mx/?760a7b3ddd30df3d#/riYuiG/+LWrvUgH7bDLfeXYfmtq/WePTYbQwfAXLaU=)

gauteh:
Thanks, you have lynx installed? Also just so you know the content of the
html is in the log in case it is private.

ons. 3. apr. 2019 kl. 17:57 skrev Max <notifications@github.com>:

gauteh:
Please try the latest version again, moved to w3m which seems to be a little bit more robust. But maybe not as nice output.

jimcheetham:
This is working for me (once I realised that my startup options were being *invisibly ignored* because there was an existing astroid session running in a different desktop).

Also, because you're execve() the call, I can successfully invoke an arbitrary pipeline by taking responsibility for using "bash -c 'cmd | cmd'" if I want to.

I would *love* to be able to use something like this on message preview, instead of having to explicitly invoke my preview option ...

gauteh:
Great! What do you mean with preview and preview option?

jimcheetham:
As per my original #319; but where I originally said that "Not having them render HTML by default seems sensible" I'm sort of less comfortable with that over time. Possibly not the right conversation to have on a patch that addresses the reply function though.

gauteh:
You can choose to have the HTML part to be shown by default, think it is `thread_view.preferred_type` or `thread_view.preferred_html_only` (the latter opens html parts of HTML only part, otherwise uses the text part). Just to make sure, you do know you can activate/press enter on the html part to open it?

gauteh:
Something that would be nice would be to also include attachments (like images) and have the CID be updated accordingly (e.g. #597). But might not be any point without converting to markdown.

jimcheetham:
Thanks, I'll check later. Yes, I can currently read messages that have no text/plain portion by selecting the text/html part, and pressing enter, but I hadn't looked for a 'display by default' option recently, so I'll double-check.

gauteh:
@mxmehl: did you have a chance to try again?

mxmehl:
Yes, I tried it today and it works! w3m made the difference, lynx dump only worked with very simple (and valid) HTML, for HTML-only newsletters which tend to contain a lot of inline CSS and invalid HTML, it seems to be unstable.

gauteh:
Nice, what error did you get with lynx -dump -stdin?

mxmehl:
> Nice, what error did you get with lynx -dump -stdin?

None. Nothing happened in the UI, and the log looked the same as in my first comment. It only worked for simple (valid?) HTML. I can forward you one of the failing emails if you want

gauteh:
Ok, I think I have one message in the test suite where I get the same error. Looks like a conversion error when lynx fails.. either way we should handle whatever output the processor throws at us.

gauteh:
Merged. Not working for lynx, but seems to work well with w3m. Ideally need to fix conversion error at later time.

l1gi:
I have hit probably the same issue.

Log file: https://cloud.tekkirk.org/index.php/s/dJHM6NbLZtc597T

> [17:33:15] [0x00000008099b5000] [M] [error] astroid needs an UTF-8 locale! this is probably not going to work.
> ...
> [17:33:23] [0x00000008099b5000] [M] [info] cmd: running: w3m -dump -T text/html
> [17:33:23] [0x00000008099b5000] [M] [debug] em: deconstruct.
> [17:33:23] [0x00000008099b5000] [M] [info] db: closing db.
> [17:33:23] [0x00000008099b5000] [M] [debug] db: ro: waiting for lock to close..
> [17:33:23] [0x00000008099b5000] [M] [debug] db: ro: closing..
> [17:33:23] [0x00000008099b5000] [M] [debug] mt: destruct.
> [17:33:23] [0x00000008099b5000] [M] [debug] ms: deconstruct
> [17:33:23] [0x00000008099b5000] [M] [debug] chunk: deconstruct.
> 
> (<unknown>:44954): glibmm-CRITICAL **: 17:33:23.538: 
> unhandled exception (type Glib::Error) in signal handler:
> domain: g_convert_error
> code  : 1
> what  : Invalid byte sequence in conversion input

Mail: https://cloud.tekkirk.org/index.php/s/n5oSGPi2PfwjYzY

gauteh:
See the first quoted line, can you try with an utf-8 locale?


ons. 14. aug. 2019, 17:41 skrev l1gi <notifications@github.com>:

l1gi:
Yes, that was the reason why I pasted it. My previous locale setting:

```
LANG=
LC_CTYPE="en_US.UTF-8"
LC_COLLATE="en_US.UTF-8"
LC_TIME="en_US.UTF-8"
LC_NUMERIC="en_US.UTF-8"
LC_MONETARY="en_US.UTF-8"
LC_MESSAGES="en_US.UTF-8"
LC_ALL=en_US.UTF-8
```

After I have set LANG also to en_US.UTF-8 it started to work. I thought my setting is valid, so maybe it is not.

Thanks much for your time.

gauteh:
Ok, I tried to rely on glib there so I hope it is correctly interpreting
your env. The settings might affect the output from the HTML to text
processor for quoting.

tor. 15. aug. 2019, 09:56 skrev l1gi <notifications@github.com>:

Issue #636, opened 2019-05-24
Segmentation fault or keeps running in background when quitting


Liblor:
Hello,

If I quit (pressing `q` and confirming with `y`) the program either crashes or only the gui goes away but it keeps running in the background.

The backtrace of the segfault:

```
           PID: 25441 (astroid)
           UID: 1000 (loris)
           GID: 1000 (loris)
        Signal: 11 (SEGV)
     Timestamp: Fri 2019-05-24 11:10:34 CEST (34min ago)
  Command Line: astroid
    Executable: /usr/local/bin/astroid
 Control Group: /user.slice/user-1000.slice/session-2.scope
          Unit: session-2.scope
         Slice: user-1000.slice
       Session: 2
     Owner UID: 1000 (loris)
       Boot ID: e86f26f8322f44f9b26652fef8e4a993
    Machine ID: 8b2d02ccdcb74d23a5a01cb3af2d87ed
      Hostname: localhost
       Storage: /var/lib/systemd/coredump/core.astroid.1000.e86f26f8322f44f9b26652fef8e4a993.25441.1558689034000000.lz4
       Message: Process 25441 (astroid) of user 1000 dumped core.
                
                Stack trace of thread 25441:
                #0  0x00007f18b6f42e09 boost::log::v2_mt_posix::core::remove_all_sinks() (libboost_log.so.1.69.0)
                #1  0x00000000004cf346 Astroid::Astroid::~Astroid() (astroid)
                #2  0x00000000004cf549 Astroid::Astroid::~Astroid() (astroid)
                #3  0x00007f18b8120f37 g_datalist_clear (libglib-2.0.so.0)
                #4  0x00007f18b822ccf0 g_object_unref (libgobject-2.0.so.0)
                #5  0x00007f18b64fd6c0 __run_exit_handlers (libc.so.6)
                #6  0x00007f18b64fd800 exit (libc.so.6)
                #7  0x00007f18b64e6f3a __libc_start_main (libc.so.6)
                #8  0x00000000004ce78e _start (astroid)
                
                Stack trace of thread 25448:
                #0  0x00007f18b65b33f7 __poll (libc.so.6)
                #1  0x00007f18b81402de n/a (libglib-2.0.so.0)
                #2  0x00007f18b81406b3 g_main_loop_run (libglib-2.0.so.0)
                #3  0x00007f18b985bb6a n/a (libgio-2.0.so.0)
[...]

PR #639, opened 2019-05-30
Fix segmentation fault when quitting (#636)

Liblor:
I added a redundant `include` to the main for readability, I hope that's fine for you.

### DIFF
--- a/src/astroid.cc
+++ b/src/astroid.cc
@@ -418,8 +418,6 @@ namespace Astroid {
       actions->close ();
       delete actions;
     }
-
-    logging::core::get()->remove_all_sinks ();
   }
 
   int Astroid::on_command_line (const refptr<Gio::ApplicationCommandLine> & cmd) {
--- a/src/main.cc
+++ b/src/main.cc
@@ -1,7 +1,12 @@
+# include <boost/log/core.hpp>
 # include "astroid.hh"
+
+namespace logging = boost::log;
 
 int main (int argc, char **argv) {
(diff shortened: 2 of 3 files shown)


Liblor:
If I understand correctly you try to avoid stuff like:
```
6: the current user preferred locale is: en_US.utf8
6: the current user preferred locale is: en_US.utf8
6: the current user preferred locale is: en_US.utf8
```
unfortunately it doesn't look like this patch helps. I diffed the output of `ctest --verbose` and didn't see any noticeable differences

gauteh:
On Thu, May 30, 2019 at 11:28 PM Loris Reiff <notifications@github.com>
wrote:

> If I understand correctly you try to avoid stuff like:
>
> 6: the current user preferred locale is: en_US.utf8
> 6: the current user preferred locale is: en_US.utf8
> 6: the current user preferred locale is: en_US.utf8
>
> unfortunately it doesn't look like this patch helps. I diffed the output
> of ctest --verbose and didn't see any noticeable differences
>
Hm, weird, it worked here. Compared to master or compared to your PR? I
believe on master the duplicate lines should be gone. If I remember
correctly that was the purpose of the original commit.

Liblor:
I just had a look at my bash history, I compared the patched PR to master, of course there's no difference... I guess I was a little tired. Sorry about the inconvenience... The above mentioned duplicated lines are also present on master.

gauteh:
Thanks, merged!

Issue #638, opened 2019-05-28
gpg.always_trust doesnt work with gmime 3..


emdete:
always trust isnt set in code anymore

gauteh:
you have to set it in the gpg config with new gmime

emdete:
The context doesnt take that flag anymore but the encrypt call itself (src/crypto.cc:246), so the line 251 should read:

```
always_trust?GMIME_ENCRYPT_ALWAYS_TRUST:GMIME_ENCRYPT_NONE,
```

do you agree?

gauteh:
That seems right

emdete:
do you want me to create a PR or is that line enough?

gauteh:
Pr would be great👍

tir. 28. mai 2019 kl. 18:51 skrev M. Dietrich <notifications@github.com>:

igoralmeida:
Tested and it works. Maybe close #637 as dupe?

gauteh:
Fixed.

PR #644, opened 2019-07-11
Fix encryption error with gmime3

mxmehl:
This makes `gpg.always_trust` work again with gmime3. Solution taken from issue #638

I consider this worth to trigger a new version or at least a backport to the latest release since it seems that gmime3 is now installed by default on Arch. At least I did not conciously install gmime3 on my laptop, and after a recent rebuild of astroid encryption suddenly failed with "Failed encrypting: Encryption failed: Unusable public key".

### DIFF
--- a/src/crypto.cc
+++ b/src/crypto.cc
@@ -248,7 +248,7 @@ namespace Astroid {
         mo,
         sign,
         userid.c_str (),
-        GMIME_ENCRYPT_NONE,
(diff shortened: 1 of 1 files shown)


mxmehl:
Just investigated the situation on Arch: notmuch 0.29.1-1 started to use gmime3 as a dependency, so that's the reason why astroid starts using gmime3.

gauteh:
Thanks, merged (fixed formatting).

Issue #646, opened 2019-07-30
SIGSEGV on FreeBSD during attempt to use non supported abstract socket


l1gi:
Hi,

I got 2 SIGSEGVs during my first try of astroid.

First)
$ astroid --log-level trace -n
[2019-07-30 17:14:56.841653] [0x0000000809d50000] [info]    welcome to astroid! - v0.15
[2019-07-30 17:14:56.918366] [0x0000000809d50000] [info]    creating new config..
[2019-07-30 17:14:56.918409] [0x0000000809d50000] [debug]   HOME: /home/ligi
[2019-07-30 17:14:56.918603] [0x0000000809d50000] [error]   the config file: /home/ligi/.config/astroid/config already exists.
Abort trap (core dumped)
$ gdb usr/bin/astroid astroid.core 
...
Core was generated by `astroid --log-level trace -n'.
Program terminated with signal SIGABRT, Aborted.
#0  0x0000000806efc45a in thr_kill () from /lib/libc.so.7
[Current thread is 1 (LWP 101163)]
(gdb) 
(gdb) bt
#0  0x0000000806efc45a in thr_kill () at /lib/libc.so.7
#1  0x0000000806efa844 in raise () at /lib/libc.so.7
#2  0x0000000806e6d079 in abort () at /lib/libc.so.7
#3  0x00000008009fa31b in  () at /usr/local/lib/libsigc-2.0.so.0
#4  0x5563bfcf7c0706a0 in  ()
#5  0x00000008009fa73f in  () at /usr/local/lib/libsigc-2.0.so.0
#6  0x00000008009fa47a in sigc::trackable::~trackable() () at /usr/local/lib/libsigc-2.0.so.0
#7  0x0000000805f06676 in Glib::Threads::Private<Glib::DispatchNotifier>::delete_ptr(void*) ()
    at /usr/local/lib/libglibmm-2.4.so.1
#8  0x00000008041b22a2 in g_private_replace () at /usr/local/lib/libglib-2.0.so.0
#9  0x0000000805f06149 in Glib::DispatchNotifier::unreference_instance(Glib::DispatchNotifier*, Glib::Dispatcher const*) () at /usr/local/lib/libglibmm-2.4.so.1
#10 0x0000000805f063b6 in Glib::Dispatcher::~Dispatcher() () at /usr/local/lib/libglibmm-2.4.so.1
#11 0x0000000806edd0c1 in __cxa_finalize () at /lib/libc.so.7
[...]

PR #649, opened 2019-08-15
Support plaforms without abstract unix sockets

serverhiccups:
This is a fix to #646 that should allow functionality on macOS and FreeBSD. Sorry if the code style wasn't up to scratch.


gauteh:
I think this approach is good, maybe adapt the path management from #656 ?

gauteh:
Fixed in other PR.

PR #651, opened 2019-08-19
Add get_queries() extension method

glacambre:
This patch adds a way to programmatically generate queries on startup.
This is done by adding a new method named "get_queries" to the
astroid_activatable interface.

Closes https://github.com/astroidmail/astroid/issues/614.

Since I still don't know how to write C++<->python interop code, I took what I feel is a particularly hackish approach. Indeed, I made the python plugin return a `[name, query, name, query…]` list instead of a list of tuples or even a map because I couldn't figure out how to have the C++ side deal with objects or lists of lists. Any help improving this part of the code is more than welcome.

### DIFF
--- a/src/plugin/manager.cc
+++ b/src/plugin/manager.cc
@@ -241,6 +241,32 @@ namespace Astroid {
     return clrs;
   }
 
+  std::vector<std::pair<ustring, ustring>> PluginManager::AstroidExtension::get_queries () {
+    std::vector<std::pair<ustring, ustring>> queries = {};
+
+    if (!active || astroid->plugin_manager->disabled) return queries;
+
+    for (PeasPluginInfo * p : astroid->plugin_manager->astroid_plugins) {
+      PeasExtension * pe = peas_extension_set_get_extension (extensions, p);
+
+      GList * list = astroid_activatable_get_queries (ASTROID_ACTIVATABLE(pe));
+
+      if (list != NULL) {
+        std::vector<ustring> _list = Glib::ListHandler<ustring>::list_to_vector (list, Glib::OWNERSHIP_NONE);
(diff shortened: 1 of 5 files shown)


src/plugin/manager.cc
```diff
@@ -241,6 +241,30 @@ namespace Astroid {
...
+
+    if (!active || astroid->plugin_manager->disabled) return queries;
+
+    for (PeasPluginInfo * p : astroid->plugin_manager->astroid_plugins) {
+      PeasExtension * pe = peas_extension_set_get_extension (extensions, p);
+
+      GList * list = astroid_activatable_get_queries (ASTROID_ACTIVATABLE(pe));
+
+      if (list != NULL) {
+        std::vector<ustring> _list = Glib::ListHandler<ustring>::list_to_vector (list, Glib::OWNERSHIP_NONE);
```
gauteh: What happens with the ownership of `list` is it cleaned up?

gauteh:
I think this approach should be good!

src/plugin/manager.cc
glacambre: I have no idea. I think I'm doing the exact same thing as what happens in `PluginManager:AstroidExtension::get_tag_colors`, which I assumed did everything right memory-management/ownership-wise.

I just ran astroid in Valgrind and it seems my assumption was wrong, get_queries() does leak memory. This probably means that other `PluginManager::AstroidExtension::get_*` functions also leak memory (but I could be wrong, I'm pretty bad at C++).

glacambre: I added a `g_list_free(list)` call, please let me know if you think this is enough.

gauteh: Looks good, you are probably right that there is a bug i the other methods.

PR #656, opened 2019-08-27
FreeBSD and Mac does not provide abstract sockets.

l1gi:
My approach. Tested just on FreeBSD. Need Mac as well as Linux testing.

### DIFF
--- a/src/config.cc
+++ b/src/config.cc
@@ -100,6 +100,9 @@ namespace Astroid {
       std_paths.cache_dir = path(cache) / path("astroid");
     }
 
+    /* socket path */
+    std_paths.socket_dir = std_paths.cache_dir / path("socket");
+
     /* default runtime */
     char * runtime = getenv ("XDG_RUNTIME_HOME");
     if (runtime == NULL) {
@@ -336,6 +339,11 @@ namespace Astroid {
     if (!is_directory(std_paths.runtime_dir)) {
       LOG (warn) << "cf: making runtime dir..";
       create_directories (std_paths.runtime_dir);
+    }
+
+    if (!is_directory(std_paths.socket_dir)) {
+      LOG (warn) << "cf: making socket dir..";
+      create_directories (std_paths.socket_dir);
     }
 
     if (!is_regular_file (std_paths.config_file)) {
--- a/src/config.cc
+++ b/src/config.cc
@@ -1,6 +1,7 @@
 # include <iostream>
 # include <stdlib.h>
 # include <functional>
+# include <sys/stat.h>
 
 # include <boost/filesystem.hpp>
 # include <boost/filesystem/operations.hpp>
@@ -343,7 +344,9 @@ namespace Astroid {
 
     if (!is_directory(std_paths.socket_dir)) {
       LOG (warn) << "cf: making socket dir..";
+      mode_t um = umask (077);
       create_directories (std_paths.socket_dir);
+      umask (um);
     }
 
     if (!is_regular_file (std_paths.config_file)) {
--- a/src/modes/thread_view/page_client.cc
+++ b/src/modes/thread_view/page_client.cc
@@ -116,14 +116,18 @@ namespace Astroid {
     /* set up unix socket */
     LOG (warn) << "pc: id: " << id;
 
-    socket_addr = ustring::compose ("%1/sockets/astroid.%2.%3.%4",
-        astroid->standard_paths ().runtime_dir.c_str(),
+    socket_addr = ustring::compose ("%1/astroid.%2.%3.%4",
+        astroid->standard_paths ().socket_dir.c_str(),
         getpid(),
         id,
         UstringUtils::random_alphanumeric (30));
 
     refptr<Gio::UnixSocketAddress> addr = Gio::UnixSocketAddress::create (socket_addr,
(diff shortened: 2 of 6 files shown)


src/modes/thread_view/page_client.cc
```diff
@@ -116,14 +116,18 @@ namespace Astroid {
...
+    socket_addr = ustring::compose ("%1/astroid.%2.%3.%4",
+        astroid->standard_paths ().socket_dir.c_str(),
         getpid(),
         id,
         UstringUtils::random_alphanumeric (30));
 
     refptr<Gio::UnixSocketAddress> addr = Gio::UnixSocketAddress::create (socket_addr,
+# if defined __FreeBSD__ || defined __APPLE__
+        Gio::UNIX_SOCKET_ADDRESS_PATH);
+# elif
```
gauteh: should be # else

src/modes/thread_view/webextension/tvextension.cc
```diff
@@ -147,7 +147,11 @@ AstroidExtension::AstroidExtension (
   const char * caddr = g_variant_get_string ((GVariant *) gaddr, &sz);
 
   refptr<Gio::UnixSocketAddress> addr = Gio::UnixSocketAddress::create (caddr,
+# if defined __FreeBSD__ || defined __APPLE__
+      Gio::UNIX_SOCKET_ADDRESS_PATH);
+# elif
```
gauteh: Same

src/modes/thread_view/thread_view.cc
```diff
@@ -1385,12 +1385,12 @@ namespace Astroid {
 
     keys.register_key ("a",
         "thread_view.archive_thread",
-        "Toggle 'inbox' tag on the whole thread",
+        "Toggle 'archived' tag on the whole thread",
         [&] (Key) {
 
           if (!edit_mode && focused_message) {
             if (thread) {
-              main_window->actions->doit (refptr<Action>(new ToggleAction(thread, "inbox")));
```
gauteh: This seems unrelated, did it sneak in?

gauteh:
I think the way of determining whether the abstract sockets is possible done in #649 is more general, though I like the path management that you do here.

src/modes/thread_view/thread_view.cc
l1gi: Yes, sorry, I will patch it out.

l1gi:
Now it could be complete. Could you review it, please?

src/config.cc
```diff
@@ -338,6 +341,11 @@ namespace Astroid {
       create_directories (std_paths.runtime_dir);
     }
 
+    if (!is_directory(std_paths.socket_dir)) {
+      LOG (warn) << "cf: making socket dir..";
```
gauteh: Should probably be debug level

l1gi:
Now the permissions should be correct:

ligi@wand:~/workspace/astroid$ ls -ld ~/.cache/astroid/socket
drwx------  2 ligi  ligi  3 Sep  2 09:54 /home/ligi/.cache/astroid/socket
ligi@wand:~/workspace/astroid$ ls -l ~/.cache/astroid/socket
total 1
srwx------  1 ligi  ligi  0 Sep  2 09:54 astroid.17482.1.9z3wdebt6f7sxdvi6tmfrwal9nvvmg

I would keep warn level if you don't mind as all other LOG() calls about 'making' are warn as well. Any objection?

gauteh:
Makes sense.

serverhiccups:
This looks like this supersedes my pr #649 so I'll close mine.

gauteh:
Thanks, merged.

gauteh:
Hi, there are some issues which seem to be related to these changes: #661.
Do you have the chance to investigate?

ons. 4. sep. 2019, 07:12 skrev serverhiccups <notifications@github.com>:

l1gi:
Hi Gaute,

unfortunately, I don't have a chance to try on different platform than 
FreeBSD.

Could be that socket bind() is called more that once without previous 
unbinding/closing?

Also, could anyone on Linux verify that proper code path is taken when 
socket type is selected, please?

Sorry for more questions than answers.

Regards
-- 
m.

gauteh:
Hi, unfortunately I do not have time to look into this in the immediate future, so the alternative is to revert the relevant commits for the time being. If you find time to handle this in some way, then that would very good, but I understand if you do not.

src/modes/thread_view/page_client.cc
```diff
@@ -116,14 +116,20 @@ namespace Astroid {
...
-        astroid->standard_paths ().runtime_dir.c_str(),
-        getpid(),
-        id,
-        UstringUtils::random_alphanumeric (30));
-
-    refptr<Gio::UnixSocketAddress> addr = Gio::UnixSocketAddress::create (socket_addr,
-        Gio::UNIX_SOCKET_ADDRESS_ABSTRACT);
+    refptr<Gio::UnixSocketAddress> addr;
+    if(Gio::UnixSocketAddress::abstract_names_supported ()) {
+      addr = Gio::UnixSocketAddress::create (socket_addr,
```
kyrias: #661 is caused by `socket_addr` being uninitialized here.

PR #663, opened 2019-10-04
Set black text colour in email body

### LINKED ISSUE #662: White text on white background in text emails with dark theme
As shown in the screenshots below, Astroid uses the text colour specified in the current GTK theme to render plain text emails, making them unreadable when using a dark theme. Note that HTML emails are rendered as expected.

This is puzzling, since the default `thread-view.scss` file (which I did not override) contains the following line:

```
.email {
    ...
    color: black;
    ...
}
```

astroid_dark_theme
astroid_light_theme

donbex:
`part.scss` sets the background colour of email bodies to be white, but didn't specify a foreground colour, causing emails to be unreadable when using a dark GTK theme. Fixes #662.

### DIFF
--- a/ui/part.scss
+++ b/ui/part.scss
@@ -18,6 +18,7 @@
 
 body {
   background-color: white !important;
+  color: black;
   overflow-x: auto;
   overflow-y: hidden;
   word-break: break-word;


gauteh:
Thanks, merged!

PR #664, opened 2019-10-06
page_client: Initialize socket_addr when creating an abstract socket as well

kyrias:


gauteh:
Thanks, this looks good.

Issue #667, opened 2019-10-30
Too high minimum width for the window


bjesus:
Hi again,
I've noticed I can't resize Astroid to take just one half of my screen (or less). It seemed like some hard-coded minimum width that's specified somewhere. I tracked it down to this line:
https://github.com/astroidmail/astroid/blob/35660d486bed3c856603dcff8e8b52153e8e8da9/src/command_bar.cc#L30

After changing it to 300, it seems to work just fine. But I'm wondering, why is there a a minimum width setting in first place? What's the use of it?

Thank you!

gauteh:
I think it was just to make it look good by default. But there's probably
another setting which doesn't prevent it from being lower. PR welcome :)

bjesus:
Done!

PR #668, opened 2019-11-02
remove minimum width for command bar

bjesus:
Following on #667, I suggest removing this minimum width setting completely as Astroid already has a default size defined here:
https://github.com/astroidmail/astroid/blob/35660d486bed3c856603dcff8e8b52153e8e8da9/src/main_window.cc#L105

As a reference I found this: https://valadoc.org/gtk+-3.0/Gtk.Widget.set_size_request.html
> In most cases, set_default_size is a better choice for toplevel windows than this function; setting the default size will still allow users to shrink the window.

Thank you!

### DIFF
--- a/src/command_bar.cc
+++ b/src/command_bar.cc
@@ -27,7 +27,6 @@ namespace Astroid {
 
     hbox.pack_start (mode_label, false, false, 5);
     hbox.pack_start (entry, true, true, 5);
-    entry.set_size_request (600, -1);
     add(hbox);
 
     entry.signal_activate ().connect (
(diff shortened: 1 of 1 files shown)


gauteh:
Does this cause the entry to be shorter at the same window width? Will it be short even with a wide window?

bjesus:
You were right, removing the line did make the command bar shrink to its default width which looked too small. I've added `entry.set_max_width_chars(600);` and now the command bar stretches to fill the available area, like so:
Screenshot from 2019-11-02 17-20-45

gauteh:
Thanks, merged!

PR #672, opened 2019-12-04
Use colors with the correct gtk state flag

eonpatapon:
This applies the correct default font color for selected items (white on blue instead of black on blue) which make them more readable.

Maybe the subject colors could be also handled this way (didn't touch them since they are configurable).


gauteh:
Thanks, merging!

PR #677, opened 2020-03-25
only use plain part if possible

gauteh:


ff2000:
I remember that I got mails in the past where the content was in the HTML part and the plain part only contained a message that the content couldn't be displayed. Likely those were just info mails, and the fact that I didn't come across those since quite a while makes this a non-issue. But I still wanted to mention it.

gauteh:
In that case the alternative part should be selectable and then expandable?

gauteh:
I do very occasionally come across those messages, but since you can open those manually, this feels like the most intuitive behavior.

Issue #676, opened 2020-02-23
3 testsuite errors: message reading produces the same output as compose message input


jonassmedegaard:
recent builds on Debian fails testsuite at 3 places, with identical error message:
> message reading produces the same output as compose message input

More detail in Debian bugreport at https://bugs.debian.org/952167
Full build log is at http://qa-logs.debian.net/2020/02/22/astroidmail_0.15-3_unstable.log and should also appear soon at https://buildd.debian.org/status/package.php?p=astroidmail&suite=experimental

jonassmedegaard:
bump

stefanor:
Took a look into this. In Debian regression started with gmime 3.2.4, the tests passed with gmime 3.2.1.

The offending change appears to be jstedfast/gmime@2dd1be81555d787ff1ff03b11e5cb4b9e81630d6.

Putting a trailing newline on the test bodies avoids this test failure.

I think it's reasonable to append a trailing newline to a `text/plain` body. I notice that Python's MIME implementation will do this, too.

jonassmedegaard:
Quoting Stefano Rivera (2020-05-09 03:49:22)
> Took a look into this. In Debian regression started with gmime 3.2.4, the tests passed with gmime 3.2.1.
> 
> The offending change appears to be jstedfast/gmime@2dd1be81555d787ff1ff03b11e5cb4b9e81630d6.
> 
> Putting a trailing newline on the test bodies avoids this test failure.
> 
> I think it's reasonable to append a trailing newline to a `text/plain` body. I notice that Python's MIME implementation will do this, too.

Great. Thanks!

-- 
 * Jonas Smedegaard - idealist & Internet-arkitekt
 * Tlf.: +45 40843136  Website: http://dr.jones.dk/

 [x] quote me freely  [ ] ask before reusing  [ ] keep private

PR #685, opened 2020-05-09
End test email message bodies with a trailing newline

stefanor:
Test regressed under gmime 3.2.3, because of the change in jstedfast/gmime@2dd1be81555d787ff1ff03b11e5cb4b9e81630d6

A trailing newline was added to the text/plain message body in the test, causing the round-trip comparison to fail.

Fixes: #676

### DIFF
--- a/tests/test_composed_message.cc
+++ b/tests/test_composed_message.cc
@@ -32,7 +32,7 @@ BOOST_AUTO_TEST_SUITE(Composing)
 
     ComposeMessage * c = new ComposeMessage ();
 
-    ustring bdy = "This is test: æøå.\n > testing\ntesting\n...";
+    ustring bdy = "This is test: æøå.\n > testing\ntesting\n...\n";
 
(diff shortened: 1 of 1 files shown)


gauteh:
Stefano Rivera writes on May 9, 2020 3:54:
> Test regressed under gmime 3.2.3, because of the change in
> jstedfast/gmime@2dd1be81555d787ff1ff03b11e5cb4b9e81630d6
>
> A trailing newline was added to the text/plain message body in the test,
> causing the round-trip comparison to fail.

Won't this fail on other versions of gmime? Is this part of the newest
gmime?

Regards, Gaute

stefanor:
> Won't this fail on other versions of gmime?

Nope. gmime doesn't seem to have been doing any stripping there in the past, it just started to add a trailing newline.

As you can see from the CI tests, it works across at least gmime 3.2.0 (Ubuntu bionic) to 3.2.7 (Debian unstable & Ubuntu devel)

gauteh:
Ok, thanks, merged

PR #686, opened 2020-05-09
GitHub action to build on Debian & Ubuntu

stefanor:
I'd like to think #676 could have been caught sooner, if there was continuous integration testing. And presumably other changes that have caused build failures in master could have been caught sooner.

Here's a [GitHub Actions](https://help.github.com/en/actions) config to run builds on a range of Debian and Ubuntu releases. You could easily add other distributions using the same approach.

This will need some maintenance over time:
1. Updating the list of releases to test on. Ubuntu devel points to the current development release, but the other supported LTSs are named.
1. Updating build dependencies. This might require hacks like the `apt-get ... || apt-get ...` I used here. In the future `apt-get satisfy` will be a nice solution for that kind of thing, but it's only available from apt 1.9.0.

You can see logs of this build [here](https://github.com/stefanor/astroid/actions/runs/99741389). They're fail on newer releases due to #676, but [here](https://github.com/stefanor/astroid/actions/runs/99741390) is an example with #685 merged, so they all pass.

I started with Travis CI: [Branch](https://github.com/stefanor/astroid/compare/travis-ci?expand=1), [Status](https://travis-ci.org/github/stefanor/astroid). But I could only test on one Ubuntu LTS release, there. Which means it'd take a lot longer to notice regressions caused by changes in dependency libraries. Testing on Debian unstable / Ubuntu devel releases is a good way to catch these sooner.

### DIFF
--- a/.github/workflows/ci-debian-build-test.yml
+++ b/.github/workflows/ci-debian-build-test.yml
@@ -1,0 +1,72 @@
+---
+name: "CI: Debian & Derivatives"
+on:
+  - push
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    container: ${{ matrix.container }}
+    strategy:
+      fail-fast: false
+      matrix:
+        container:
+          - 'debian:sid'
+          - 'debian:stable'
+          - 'ubuntu:devel'
+          - 'ubuntu:focal'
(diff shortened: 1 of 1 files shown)


gauteh:
Thanks, merging.

Issue #695, opened 2020-07-21
Crash: Selecting multiple threads and then pressing Enter on one


matthiasbeyer:
I've just experienced a crash:

Select multiple elements and then hit Enter on one.

matthiasbeyer:
Ah, it seems that the crash is not related to that I selected multiple entries.

matthiasbeyer:
Maybe this helps a bit:

```
(.astroid-wrapped:1826): glibmm-ERROR **: 16:57:40.701: 
unhandled exception (type std::exception) in signal handler:
what: db: got more than one thread for thread id.

Trace/breakpoint trap
```

matthiasbeyer:
Some more:

```
(gdb) bt
#0  0x00007ffff2a18c75 in _g_log_abort () from /nix/store/hp2zfbm7avn7xkbxsrh3zhzwzpw7jsvd-glib-2.62.4/lib/libglib-2.0.so.0
#1  0x00007ffff2a19ce5 in g_log_default_handler () from /nix/store/hp2zfbm7avn7xkbxsrh3zhzwzpw7jsvd-glib-2.62.4/lib/libglib-2.0.so.0
#2  0x00007ffff2a19f1b in g_logv () from /nix/store/hp2zfbm7avn7xkbxsrh3zhzwzpw7jsvd-glib-2.62.4/lib/libglib-2.0.so.0
#3  0x00007ffff2a1a0ff in g_log () from /nix/store/hp2zfbm7avn7xkbxsrh3zhzwzpw7jsvd-glib-2.62.4/lib/libglib-2.0.so.0
#4  0x00007ffff20d8db3 in Glib::exception_handlers_invoke() [clone .cold] () from /nix/store/hb7vprqk1j49z321lkigmxy7a5abazh0-glibmm-2.62.0/lib/libglibmm-2.4.so.1
#5  0x00007ffff273643e in Gtk::Widget_Class::key_press_event_callback(_GtkWidget*, _GdkEventKey*) [clone .cold] () from /nix/store/qlpp5c4qryb0dya6mlprkqiqkj1h8bh5-gtkmm-3.24.2/lib/libgtkmm-3.0.so.1
#6  0x00007ffff48e96cb in _gtk_marshal_BOOLEAN__BOXEDv () from /nix/store/4y5ird78yicid80g95iq1hzmshb7xjv9-gtk+3-3.24.13/lib/libgtk-3.so.0
[...]

PR #696, opened 2020-07-22
Fix segfault on opening multiple threads for one entry in thread list

matthiasbeyer:
This is my take on #695 

I'm not a C++ dev though (mainly I'm a Rust dev), thus I guess this is something one might not want to have here.

For my case, it fixes the crash when opening a thread where notmuch returns multiple threads for one thread id.
Please have a look.

### DIFF
--- a/src/db.cc
+++ b/src/db.cc
@@ -451,7 +451,6 @@ namespace Astroid {
     string query_s = "thread:" + thread_id;
 
     notmuch_query_t * query = notmuch_query_create (nm_db, query_s.c_str());
-    notmuch_thread_t  * nm_thread;
     notmuch_threads_t * nm_threads;
     notmuch_status_t st = NOTMUCH_STATUS_SUCCESS;
 
@@ -469,13 +468,8 @@ namespace Astroid {
     int c = 0;
     for ( ; notmuch_threads_valid (nm_threads);
          notmuch_threads_move_to_next (nm_threads)) {
-      if (c > 0) {
-        LOG (error) << "db: got more than one thread for thread id.";
-        throw invalid_argument ("db: got more than one thread for thread id.");
-      }
-
-      nm_thread = notmuch_threads_get (nm_threads);
-
+      notmuch_thread_t* nm_thread = notmuch_threads_get (nm_threads);
+      func (nm_thread);
       c++;
     }
 
@@ -485,9 +479,6 @@ namespace Astroid {
       notmuch_query_destroy (query);
       return;
     }
-
-    /* call function */
-    func (nm_thread);
 
     /* free resources */
     notmuch_query_destroy (query);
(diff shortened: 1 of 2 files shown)


matthiasbeyer:
Note that this does _not_ open multiple new panes, one for each thread returned by notmuch. I would have expected that behaviour, but it does not do that. I don't know why.

gauteh:
Thanks, Only one additional pane is show

gauteh:
on_thread is used several places, not sure if it is safe to apply the function to all threads. E.g. a user-defined hook to delete a thread. Why does notmuch return several threads? Is it a notmuch bug? Maybe it is safer to work on the first thread if we cannot avoid the underlying issue?

matthiasbeyer:
TBH I don't know. But I've experienced this before in notmuch. As far as I can see, nothing (in the notmuch docs) says that for a `thread:<thread id>` query, notmuch does only return one thread.

So I guess it is actually expected to return multiple thread and the bug is on our side here when we expect only one thread to be returned.

gauteh:
Thread IDs should be unique. I think I've discussed this on the list
before. Maybe check with notmuch list or with #notmuch?

ons. 22. jul. 2020, 15:37 skrev Matthias Beyer <notifications@github.com>:

matthiasbeyer:
I found this: https://notmuchmail.org/pipermail/notmuch/2018/026381.html

gauteh:
I think that thread was promoted by an astroid issue. Maybe we should pick
up the issue again with notmuch? Does the threads always contain the same
messages?

ons. 22. jul. 2020, 15:40 skrev Matthias Beyer <notifications@github.com>:

matthiasbeyer:
> Does the threads always contain the same messages?

Not sure what you mean by that.

From what I see in the codebase right now, `Db::on_thread` is not used too extensively. Maybe it would be worth a try to check all cases and rewrite it to not expect there be only one thread anymore?

gauteh:
Hm, I think it is an error from the notmuch lib. We should bail or try to
recover. Executing the same action, intended to be done on a single unique
thread, multiple times is not expected by the rest of astroid or the user
and could easily cause further bugs down the line.

ons. 22. jul. 2020, 15:46 skrev Matthias Beyer <notifications@github.com>:

mreppen:
I have solved this problem by running `notmuch reindex` on the first of such split threads. This corrects it to one thread. Is there a reason this could not be attempted by astroid as a convenience to the user? Or does reindexing potentially cause problems?

matclab:
> I have solved this problem by running `notmuch reindex` on the first of such split threads. This corrects it to one thread. Is there a reason this could not be attempted by astroid as a convenience to the user? Or does reindexing potentially cause problems?

Thanks it has solved a similar problem here.

PR #698, opened 2020-08-07
ql: Default initialize threads to NULL.

ibuclaw:
The function notmuch_query_search_threads() only [sets the pointer on success](https://github.com/notmuch/notmuch/blob/2b62ca2e3b786beca8d89fa737bda0b49faa638d/lib/query.cc#L532-L533).  Not initializing the pointer could result in an abort or segfault in notmuch_threads_destroy() due to it pointing to whatever was previously on the stack.

### DIFF
--- a/src/modes/thread_index/query_loader.cc
+++ b/src/modes/thread_index/query_loader.cc
@@ -123,7 +123,7 @@ namespace Astroid {
     if (!in_destructor) stats_ready.emit ();
 
     /* set up query */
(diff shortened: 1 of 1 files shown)


gauteh:
Thanks, merged.

ibuclaw:
Cheers.

Forgot to mention to reproduce the segv, just do a failed search term twice, e.g:

```
subject:/bad: /
```

PR #700, opened 2020-09-16
use std::placeholders everywhere

Hello71:
before this was silently relying on boost placeholders, but those
default off now.


gauteh:
Thanks, merged.

PR #713, opened 2021-08-09
Use native WebKit on macOS

c-alpha:
WebKitGTK is a nightmare to build on macOS. It takes hours to compile (literally), it needs patches, and it doubles up functionality that is already available on the platform.

Hence, it would seem tempting to try to use the native WebKit that ships with the macOS platform. This is what this PR attempts.


c-alpha:
`thread_view.cc` compiles without errors, while `thread_view.hh` is free from any webkit references. So far. so good.

c-alpha:
Further files with webkit references are:

- `modes/thread_view/webextension/tvextension.cc`
- `modes/thread_view/webextension/tvextension.hh`
- `modes/thread_view/webextension/dom_utils.cc`
- `modes/thread_view/webextension/dom_utils.hh`
- `modes/thread_view/page_client.cc`
- `modes/thread_view/page_client.hh`
- `plugin/thread_view_activatable.c`

The next logical target would seem `page_client.cc|hh`, since it is a `friend` class of `ThreadView`, and it is also the next translation unit from the above list in the build process.

PR #716, opened 2022-03-24
Use symbolic icons (fixes #715)

### LINKED ISSUE #715: sending mail hangs astroid with adwaita-icon-theme 42 installed
Attempting to send mail causes the following error to be printed:

```
** (astroid:66541): CRITICAL **: 00:29:09.255:
unhandled exception (type Glib::Error) in signal handler:
domain: gtk-icon-theme-error-quark
code  : 0
what  : Icon 'mail-send' not present in theme Adwaita
```

The mail is not sent, and astroid hangs displaying "sending message..". Pressing "q" or other key bindings has no effect; the program must be killed. Downgrading to adwaita-icon-theme 41 resolves the issue.

Hello71:
In adwaita-icon-theme 42, the non-symbolic versions were removed from
the standard distribution. Therefore, switch to the symbolic versions.

I don't know where gtk-apply comes from, so I didn't change that.


gauteh:
Thanks, merged!

PR #734, opened 2023-07-18
build with protobuf 23.4-1

mklinik:
Arch gave me this version of protobuf in a recent system update.

Based on https://gitlab.archlinux.org/archlinux/packaging/packages/astroid/-/blob/main/protobuf-23.patch

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -117,7 +117,9 @@ find_package ( Boost REQUIRED
   system
   )
 
+find_package (Protobuf CONFIG REQUIRED)
 find_package (Protobuf 3.0 REQUIRED)
+set (PROTOBUF_LIBRARIES protobuf::libprotobuf)
 set (PROTO_FILES
   src/modes/thread_view/webextension/messages.proto
   )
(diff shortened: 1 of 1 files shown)


jorsn:
Thanks! This seems necessary for newer protobuf in general (not only on arch), and it also still builds with protobuf 21.12.
But why change to c++17?

mklinik:
> But why change to c++17?

When I try to compile with c++14 I get the following compile error, among many others

```
In file included from /usr/include/google/protobuf/stubs/common.h:20,
                 from /usr/include/google/protobuf/io/coded_stream.h:107,
                 from /home/mkl/sandbox/software/astroid/build/messages.pb.h:26,
                 from /home/mkl/sandbox/software/astroid/src/modes/thread_view/page_client.hh:14,
                 from /home/mkl/sandbox/software/astroid/src/modes/thread_view/thread_view.cc:19:
/usr/include/absl/strings/string_view.h:53:26: error: ‘string_view’ in namespace ‘std’ does not name a type
```

Searching for this error message, I found people saying that string_view is apparently a feature introduced in c++17.

jorsn:
Ah, yes, you're right.

PR #738, opened 2023-09-09
Fix help message breaking screen layout

rothn:
Currently, this opens an empty sidebar on the left part of the window and displays a cut-off help message in a ribbon at the bottom of the window. The sidebar does not go away until exiting and re-opening the app. This change removes this behavior to improve usability while keeping the original intent in the code to track the feature for future contributors see the contribution opportunity.

I would also be open to deleting these lines.

### DIFF
--- a/src/main_window.cc
+++ b/src/main_window.cc
@@ -788,8 +788,10 @@ namespace Astroid {
     multi_waiting = true;
     multi_keybindings = kb;
 
+#if 0
     rev_multi->set_reveal_child (true);
     label_multi->set_markup (kb.short_help ());
+#endif
 
     return true;
   }


gauteh:
Isn't this going to show markup tags in the text?

ibuclaw:
This change breaks the bottom bar from showing when pressing `;`

Before:
image

After:
image


Which makes really confusing whether or not one is in "Apply action" to multiple threads mode.

rothn:
I use this on a phone, and it worked there. Hmm... If this breaks desktop users, perhaps we should revert?

ibuclaw:
> I use this on a phone, and it worked there. Hmm... If this breaks desktop users, perhaps we should revert?

I would be keen to know how you are running Astroid on a phone so I can check this.

Yes, it breaks desktop users, so I've raised the revert here --> #751

rothn:
> > I use this on a phone, and it worked there. Hmm... If this breaks desktop users, perhaps we should revert?
> 
> I would be keen to know how you are running Astroid on a phone so I can check this.
> 
> Yes, it breaks desktop users, so I've raised the revert here --> #751

Just install it on a PinePhonePro and use with the onscreen keyboard in portrait mode.

jorsn:
@rothn: Can you post a screenshot of the breakage on the pinephone? Maybe one can fix it for both phones and desktops (or: small and large displays). Otherwise, I would also prefer desktops and thus merge #751.

rothn:
My PPP doesn't work anymore and I don't have time soon to restore it or plans to purchase another one. Since this is probably not phone-specific, but rather layout-specific, you can probably repro this by running in a window of size 720x1440, in vertical/portrait mode.

ibuclaw:
The only thing I notice - if I were to resize the window to the smallest it'll allow me - is that the minimum width increases, and can't be reversed.

Smallest possible window after opening.
image

Smallest possible window after opening the `;` bar.
image

After closing the `;` bar, can't resize the window any smaller than this.
image

ibuclaw:
Possible patch.

```patch
--- a/src/main_window.cc
+++ b/src/main_window.cc
@@ -162,6 +162,7 @@ namespace Astroid {
     label_multi = Gtk::manage (new Gtk::Label ());
     rh_->pack_start (*label_multi, true, true, 5);
     label_multi->set_halign (Gtk::ALIGN_START);
+    label_multi->set_line_wrap();
 
     rev_multi->set_margin_top (0);
     rh->set_margin_bottom (5);
```

image

ibuclaw:
> The sidebar does not go away until exiting and re-opening the app.

FAOD - the only key that will close this bar is Escape - which likely doesn't exist on mobiles.

https://github.com/astroidmail/astroid/blob/327d107bb39d81ca673113621a59a2bce0754c80/src/main_window.cc#L683-L695

PR #739, opened 2023-09-09
Add a '#' key to send a message to the trash.

rothn:
Currently, Astroid has no support for deleting messages. Without deleting messages, a mail server with limited storage will eventually become full and incapable of receiving any more mail. Message deletion functionality in clients is an important mechanism for preventing this.

This adds a mechanism to mark messages as 'trash' in Astroid, which servers like gmail are configured to interpret as a deletion request.

### DIFF
--- a/src/db.cc
+++ b/src/db.cc
@@ -31,7 +31,7 @@ namespace Astroid {
 
   /* static settings */
   bool Db::maildir_synchronize_flags = false;
-  std::vector<ustring> Db::excluded_tags = { "muted", "spam", "deleted" };
+  std::vector<ustring> Db::excluded_tags = { "muted", "spam", "deleted", "trash" };
   std::vector<ustring> Db::sent_tags = { "sent" };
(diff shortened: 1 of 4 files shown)


gauteh:
I think we can add this for `trash` which has a special meaning in IMAP and is handled to some degree by notmuch. As you probably know trash isn't cleaned up normally, but is for instance by gmail remotely eventually: https://github.com/gauteh/lieer#caveats . You can achieve the same thin for custom tags by doing this: https://github.com/astroidmail/astroid/wiki/User-defined-keyboard-hooks#example-toggle-custom-tag-in-thread-index

rothn:
In case anyone who uses gmail comes across this-- "trash" seems to be the more standard way to handle this. And as far as I can tell, gmail will _eventually_ delete these message. But mapping the "TRASH" system label to "Trash" (with a capital T) in my gmail helper causes Gmail to respect this better. I think the more standard way is probably the right approach for a PR, but hopefully this helps other Gmail users who stumble across this.

PR #740, opened 2023-09-09
Remove enable-frame-flattening and iframe for message body

### LINKED ISSUE #720: webkit_settings_set_enable_frame_flattening() deprecation

rothn:
This attempts to revive https://github.com/astroidmail/astroid/pull/732 which contains an extremely important usability fix. Original message from @ibuclaw below:

As of WebKitGTK 2.40, enable-frame-flattening is no longer supported, the property does nothing, resulting in all messages in the thread viewer being cut off at around 150px, making all long messages unreadable.

I initially had a look into seeing whether the iframe height could be set programmatically with `webkit_dom_html_iframe_element_set_height` from the content of the iframe via `webkit_dom_element_get_scroll_height`. However nothing worked at page load time of an email thread, the content scroll height remained at 150 throughout the entire `ThreadView_on_load_changed` pass (because iframes are rendered lazily after the main window thread has finished?  Can't think of any other reason why this was observed). The scroll height was only found to update after the thread had finished loading, and triggering a hide/show toggle on each message.

Another option would be to use the JavaScriptCore API, and have all the logic to manipulate the height of the iframe in JS. Two possible problems with this: Firstly `enable-javascript` is explicitly disabled in Astroid's WebKit settings; second, potentially all of the `webextension` code may need to be rewritten in JS - given the huge amount of deprecation warnings when building Astroid, this likely will have to be done in anger at some point in the future anyway.

I gave up going down the path of manipulating the iframe after setting `srcdoc`, and replaced the iframe with a div instead. CSS has been fixed up and theme version bumped as it would be a breaking change for all existing users that override the theme of the UI.

I can't see any concernable difference between before and after when viewing both text/html and text/plain messages, though it is difficult to gauge as my starting point is broken messages.

The `AstroidExtension::reload_images` code path is the least tested of all the changes. The minimal check I did to see whether the logic is still fine was to send myself an email with a CID attachment, and observe it render correctly after pressing C-i in thread view.

As far as this change is concerned, this is one way to fix #720 that suits my use of Astroid. I'm not convinced that it is the correct way to go about it though, as I have no reason to dismiss the rationale for using iframes in the first place - see comments delete by this patch that make reference to style issues, deadlocks and security concerns by not having the content contained to its own iframe.

### DIFF
--- a/src/modes/thread_view/theme.hh
+++ b/src/modes/thread_view/theme.hh
@@ -27,7 +27,7 @@ namespace Astroid {
       static ustring       thread_view_css;
       static ustring       part_css;
       const char * STYLE_NAME = "STYLE";
-      const int THEME_VERSION = 5;
+      const int THEME_VERSION = 6;
 
     private:
       bool check_theme_version (bfs::path);
--- a/src/modes/thread_view/thread_view.cc
+++ b/src/modes/thread_view/thread_view.cc
@@ -99,7 +99,6 @@ namespace Astroid {
         "enable-xss-auditor", TRUE,
         "media-playback-requires-user-gesture", TRUE,
         "zoom-text-only", TRUE,
-        "enable-frame-flattening", TRUE,
 # if (DEBUG || DEBUG_WEBKIT)
         "enable-developer-extras", TRUE,
 # endif
--- a/src/modes/thread_view/webextension/tvextension.cc
+++ b/src/modes/thread_view/webextension/tvextension.cc
@@ -464,7 +464,7 @@ void AstroidExtension::handle_page (AstroidMessages::Page &s) {/*{{{*/
   webkit_dom_node_append_child (WEBKIT_DOM_NODE(head), WEBKIT_DOM_NODE(e), (err = NULL, &err));
   LOG (debug) << "done";
 
-  /* store part / iframe css for later */
+  /* store part / body_message css for later */
   part_css = s.part_css ();
 
   /* store allowed uris */
@@ -495,11 +495,9 @@ void AstroidExtension::reload_images () {
     for (auto &c : m.elements()) {
       if (!c.focusable ()) {
         WebKitDOMHTMLElement * body_container = WEBKIT_DOM_HTML_ELEMENT(webkit_dom_document_get_element_by_id (d, c.sid ().c_str ()));
-        WebKitDOMHTMLElement * iframe = DomUtils::select (WEBKIT_DOM_NODE(body_container), ".body_iframe");
-        WebKitDOMDocument * iframe_d = webkit_dom_html_iframe_element_get_content_document (WEBKIT_DOM_HTML_IFRAME_ELEMENT(iframe));
-        WebKitDOMHTMLElement * b = webkit_dom_document_get_body (iframe_d);
+        WebKitDOMHTMLElement * body_message = DomUtils::select (WEBKIT_DOM_NODE (body_container), ".body_message");
 
-        WebKitDOMNodeList * imgs = webkit_dom_element_query_selector_all (WEBKIT_DOM_ELEMENT(b), "img", (err = NULL, &err));
+        WebKitDOMNodeList * imgs = webkit_dom_element_query_selector_all (WEBKIT_DOM_ELEMENT(body_message), "img", (err = NULL, &err));
 
         gulong l = webkit_dom_node_list_get_length (imgs);
         for (gulong i = 0; i < l; i++) {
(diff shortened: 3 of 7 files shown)


gauteh:
I think that going away from the iframe the content needs to be sanitized, to avoid harmful HTML or possibly JavaScript. It could in theory re-write the entire UI and make it appear as if the email is from someone else etc. Sanitizing is difficult to get right, and critical to get right. Luckily, it is a problem that is pretty common so there should exist good tools and libraries. That is, we should not roll our own solution. If there is no C++ library it might be possible to filter it through some command.

rothn:
Hmm, yes, Geary had these issues:
* JavaScript: https://gitlab.gnome.org/GNOME/geary/-/issues/78
  * Fix: https://gitlab.gnome.org/GNOME/geary/-/merge_requests/311/diffs

A similar approach could approach for us, since Geary also did not go down the iframe path. I'm not sure what you mean by "harmful HTML" though-- could you clarify?

rothn:
Actually, it looks like Astroid takes a more aggressive security posture on Javascript than Geary by disabling it entirely: 
https://github.com/astroidmail/astroid/blob/8f2a8aeefd634c3f861ba51f7ca5fe256c7350b1/src/modes/thread_view/thread_view.cc#L88

I don't think this is an issue here. Curious as to your thoughts, @gauteh

gauteh:
It is possible to enable javascript in astroid. But that's a conscious choice. If the body text or HTML in the email body breaks the thread-view HTML it can change an entire thread. Or it can just make an enormous blank box, moving everything else out of the way so that it appears to be emails from someone else. I'm not sure if that's exactly how it works, but that's why we usually show the text part and don't render the HTML.

rothn:
> It is possible to enable javascript in astroid. But that's a conscious choice.

In that case, how would you feel about adding `"enable-javascript-markup", FALSE` to the config? If I understand correctly, this would protect the people who choose to manually enable javascript unless they explicitly allow javascript markup by editing the code.

> If the body text or HTML in the email body breaks the thread-view HTML it can change an entire thread. Or it can just make an enormous blank box, moving everything else out of the way so that it appears to be emails from someone else.

I can't easily disprove that this might be possible. However, compared to the alternative of most email being broken, and since I have not had this issue, I would suggest dealing with this if it ever comes up.

rothn:
@gauteh How would you feel about the path I suggested a couple of days ago?

gauteh:
I understand the argument, and usually I don't think "deal with it when it becomes a problem" is a deal-breaker. However, I don't think it is a good idea to ignore a known security risk, that will expose our users. There is no way back from that, and it is not ethical. We should deal with the issue in a good enough matter now, that solution probably had to be sanitize html and JavaScript. That is what makes this issue so hard to fix.

ibuclaw:
FWIW I concur with @gauteh. I still run with this patch though as for me, my email account only receives messages from text-only mailing lists, so would never hit HTML issues as pointed out both in the issue, and the code comment that gets removed by this patch.

ibuclaw:
Anyone with a little reading into this would know that Astroid wasn't the only application affected by this deprecation.

- [Gfeeds](https://gitlab.gnome.org/World/gfeeds/-/commit/77ca757f13906e5c20f84f755c4d8cbd2c729d91) just deleted the code with no apparent effect.
- [Remmina](https://gitlab.com/Remmina/Remmina/-/merge_requests/2425) ifdef'd the code out and seems like it wasn't required as well.
- [Evolution](https://gitlab.gnome.org/GNOME/evolution/-/commit/8c20556a1047b296e9ecfc8e35ba114889a0e1f6) more closely resembles astroid, they moved the logic into Javascript client render code, but then suffered [many](https://gitlab.gnome.org/GNOME/evolution/-/commit/a037d582a27ecb3bb8f51fbe0f597d9183e75701), [many](https://gitlab.gnome.org/GNOME/evolution/-/commit/feea689556fdfcbb01c7cf2428c471cfc461d459), [many](https://gitlab.gnome.org/GNOME/evolution/-/commit/974d13a5d9d3926a30c1bd601312ce9735990054), [many](https://gitlab.gnome.org/GNOME/evolution/-/commit/033db86f61bd58aff76bdf74e39d76f021726685) fallout issues from doing that.

jorsn:
> It is possible to enable javascript in astroid. But that's a conscious choice.
  -- @gauteh 

How is this possible? Only my modifying the source code? Then, I wouldn't call it configuration and hold everyone who does it responsible on their own.

gauteh:
You're right, JavaScript is disabled now (https://github.com/astroidmail/astroid/commit/09b5c4f33387941fe2965207320fb21143a14c03). Used to be possible outside the message content I think. At least we had mathjax support previously. It is still possible to mess up the view in a harmful way with just html and css.

PR #743, opened 2023-09-16
set iframe size on focus using deprecated DOM access

### LINKED ISSUE #720: webkit_settings_set_enable_frame_flattening() deprecation

jorsn:
- partly fixes #720
- alternative to removing iframes as in #740 or #732

**Problem:** iframe size is not set when a message is viewed, but only when the focus is changed after that (e.g. I open a message, it is small. Then, i type 'j'/'k' and it is correctly sized.)

**Question:** Can we call the function from some code which is called always when a message is viewed?

### DIFF
--- a/src/modes/thread_view/webextension/tvextension.cc
+++ b/src/modes/thread_view/webextension/tvextension.cc
@@ -785,6 +785,41 @@ void AstroidExtension::update_message (AstroidMessages::UpdateMessage &um) {
   ack (true);
 }
 
+void AstroidExtension::size_message_iframes () {
+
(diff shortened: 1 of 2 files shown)


gauteh:
If you want to call this when a message is focused you can use this: https://github.com/astroidmail/astroid/blob/master/src/modes/thread_view/webextension/tvextension.cc#L1781, or maybe you want https://github.com/astroidmail/astroid/blob/master/src/modes/thread_view/webextension/tvextension.cc#L1696 ?

Issue #744, opened 2024-01-25
Stop using libsoup2.4 and webkit2gtk 4.0


jbicha:
Fedora has [stopped](https://fedoraproject.org/wiki/Changes/Remove_webkit2gtk-4.0_API_Version) building the 4.0 version of webkit2gtk in preparation for the release of Fedora 40 in a few months. Debian and Ubuntu are also working to remove webkit2gtk 4.0 soon.

webkit2gtk 4.1 is the same as 4.0 except that 4.1 uses libsoup3 and 4.0 uses libsoup2.4.

There is some documentation and many examples of how to port to libsoup3 at https://gitlab.gnome.org/GNOME/libsoup/-/issues/218

PR #745, opened 2024-03-01
Use GUri instead of libsoup-2.4 and use webkit2gtk-4.1

oreo639:
This updates the URI parsing for `astroid --mailto` to use Glib instead of libsoup-2.0 (since the api was removed in libsoup-3.0).

https://libsoup.org/libsoup-3.0/migrating-from-libsoup-2.html
https://github.com/astroidmail/astroid/issues/744

I tested the mailto parsing appears to work but please double check.

### DIFF
--- a/.github/workflows/ci-debian-build-test.yml
+++ b/.github/workflows/ci-debian-build-test.yml
@@ -17,9 +17,6 @@ jobs:
           - image: 'debian:bullseye'
             env:
               WEBKITGTK_VERSION: '4.0'
-          - image: 'ubuntu:focal'
-            env:
-              WEBKITGTK_VERSION: '4.0'
           - image: 'debian:sid'
             env:
               WEBKITGTK_VERSION: '4.1'
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -95,11 +95,13 @@ endif()
 
 pkg_check_modules (GTKMM3     REQUIRED  gtkmm-3.0>=3.10)
 pkg_check_modules (GLIBMM2    REQUIRED  glibmm-2.4)
-pkg_check_modules (WEBKIT2GTK REQUIRED  webkit2gtk-4.0>=2.22)
+pkg_check_modules (WEBKIT2GTK webkit2gtk-4.1)
+if (NOT WEBKIT2GTK_FOUND)
+  pkg_check_modules (WEBKIT2GTK REQUIRED webkit2gtk-4.0>=2.22)
+endif ()
 pkg_check_modules (SASS       REQUIRED  libsass)
 pkg_check_modules (GIOMM2     REQUIRED  giomm-2.4)
-pkg_check_modules (GIOUNIX    REQUIRED  gio-unix-2.0)
-pkg_check_modules (LIBSOUP    REQUIRED  libsoup-2.4)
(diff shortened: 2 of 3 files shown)


gauteh:
Does this WebKit version work?

oreo639:
> Does this WebKit version work?

As far as I can tell, but please double check.
webkit2gtk-4.1 and webkit2gtk-4.0 are supposed to be, functionality wise and API wise, identical. (you have to explicitly opt in to -4.1 since it uses libsoup3 which cannot be loaded in to a process alongside libsoup2)

oreo639:
Would you prefer for both gui and soup2 to be supported? (e.g. `--with-soup2`)
I did it this way to avoid ifdefs, but if you are still targeting Bionic I can add that. Or would you prefer the CI be updated to target newer releases?

(e.g. if `--with-soup2` is specified, it will use `soup-2.0` and disable the ability to use `webkit2gtk-4.1` and without it, it will use guri and support both `webkit2gtk-4.1` and `webkit2gtk-4.0`)

gauteh:
No need to support bionic.

lør. 20. apr. 2024, 03:29 skrev oreo639 ***@***.***>:

jorsn:
This looks good, it works for me, with both webkit2gtk-4.1 and webkit2gtk-4.0.
It is compatible with #734 and #747.

gauteh:
I recommend moving to Ubuntu and only supporting recent distros, we don't want to spend the effort we have available on oldish libraries.

ibuclaw:
> I recommend moving to Ubuntu and only supporting recent distros, we don't want to spend the effort we have available on oldish libraries.

Agreed, EOL distributions are of no interest to anyone.

ibuclaw:
This looks to be identical to what I did as well to fix Astroid for webkit2gtk-4.1 (https://github.com/astroidmail/astroid/commit/cabdf83ec08c82f6eb44105667f62d3f287b334f), so this gets a :+1: from me.

jorsn:
I submitted the competing #748. I'd be happy if you voice your opinions about it, compared to this one.

oreo639:
Bullseye test fails because of missing G_URI_FLAGS_SCHEME_NORMALIZE, I added it because it is listed with the libsoup compatibility defines: https://gitlab.gnome.org/GNOME/libsoup/-/blob/a94a3a9a6ba14d6873c70f9931fe2d001442b14c/libsoup/soup-uri-utils.h#L40

However, for mailto links it should be useless as it only operates on these schemas: https://gitlab.gnome.org/GNOME/glib/-/blob/main/glib/guri.c#L760 https://gitlab.gnome.org/GNOME/glib/-/blob/main/glib/guri.c#L804
(e.g. it ensures that `http://example.org` becomes `http://example.org:80/`)

So I'll remove it.

jorsn:
When you change the PR, maybe you can also add the removal of ubuntu focal, since accepting this PR means dropping focal (which still has official support) from the CI.

jorsn:
I merged my PR #748 now instead of this. Thank you for this, still! #748 was born out of a closer look at this PR, here, and noticing that there was mailto handling somewhere else that did not use soup/guri. So, I unified them, which solved also #744. I might not have noticed this without this PR.

PR #747, opened 2024-05-28
set iframe height in javascript

### LINKED ISSUE #720: webkit_settings_set_enable_frame_flattening() deprecation

### LINKED ISSUE #736: Expanded mail has fixed height regardless of its content
Hi,

since some Ubuntu `apt upgrade` all expanded boxes in thread-view has fixed height approx 350px:

expanded

The only way how to see the rest of the content is to drag and move the mouse.

I removed `~/.config/astroid` directory, no change.

Collapsed view looks same:

collapsed

Could anyone tell me what could be wrong, please?

Thank you,
m.

jorsn:
- fixes #720, fixes #736
- closes #732, closes #740, closes #743

This does not allow the document (message content) to execute javascript, as can be verified with the following email, which displays "no scripts" by default, but "scripts active" when js is active.

```
Date: Tue, 1 January 1970 00:00:00 +0000
From:
Subject:
MIME-Version: 1.0
Content-Type: multipart/alternative; boundary="=-TUjG03Ah9OcLtWE56Ucm"

--=-TUjG03Ah9OcLtWE56Ucm
Content-Type: text/plain; charset=utf-8
Content-Transfer-Encoding: quoted-printable

<div id=3D"jscontent">
  no scripts
</div>

<script>
  d =3D document.getElementById('jscontent');
  d.innerHTML =3D "scripts active";
</script>

--=-TUjG03Ah9OcLtWE56Ucm
Content-Type: text/html; charset=utf-8
Content-Transfer-Encoding: quoted-printable

<!DOCTYPE html>
<html xmlns=3D"http://www.w3.org/1999/xhtml" lang xml:lang>
<head>
  <meta charset=3D"utf-8" />
  <meta name=3D"generator" content=3D"pandoc" />
  <meta name=3D"viewport" content=3D"width=3Ddevice-width, initial-scale=3D=
1.0, user-scalable=3Dyes" />
  <style>
code{white-space: pre-wrap;}
span.smallcaps{font-variant: small-caps;}
span.underline{text-decoration: underline;}
div.column{display: inline-block; vertical-align: top; width: 50%;}
</style>
</head>
<body>
<div id=3D"jscontent">
<p>no scripts</p>
</div>
<script>
  d =3D document.getElementById('jscontent');
  d.innerHTML =3D "scripts active";
</script>
</body>
</html>

--=-TUjG03Ah9OcLtWE56Ucm--
```

### DIFF
--- a/ui/thread-view.html
+++ b/ui/thread-view.html
@@ -57,9 +57,51 @@
 </div>
 
 <div id="body_template" class="body_part">
-  <iframe class="body_iframe" sandbox srcdoc=""></iframe>
+  <!--
+    sandbox="allow-same-origin" is required, otherwise the scrollHeight of the
+    embedded document cannot be accessed.
+    Do not add 'allow-scripts' to the sandbox attribute. This would allow the
+    document to execute scripts itself.
+  -->
+  <iframe class="body_iframe" sandbox="allow-same-origin" onload="handleLoadIframe(this);" srcdoc=""></iframe>
 </div>
 
+<script>
+  function onSetFocused(element, callback) {
+    const observer = new MutationObserver((mutations) => {
+      mutations.forEach((mutation) => {
+        if (mutation.target.classList.contains("focused")) {
+          callback(mutation.target);
+        }
+      });
+    });
+
+    observer.observe(element, { attributeFilter: ["class"] });
+
+    return observer.disconnect;
+  }
+
+  function sizeIframe(iframe) {
+    iframe.height = 0; // hack, no idea why it only works with this
+    iframe.height = iframe.contentWindow.document.body.scrollHeight;
+  }
+
+  function getParentWithClass(element, c) {
+    p = element.parentElement;
+    while (p != document.body && !p.classList.contains(c)) {
+      p = p.parentElement;
+    }
+
+    return p;
+  }
+
+  function handleLoadIframe(iframe) {
+    sizeIframe(iframe);
+    const email = getParentWithClass(iframe, "email");
+    if (email != document.body)
+      onSetFocused(email, _ => sizeIframe(iframe));
+  };
+</script>
 
 </body>
 </html>
--- a/src/modes/thread_view/page_client.cc
+++ b/src/modes/thread_view/page_client.cc
@@ -422,11 +422,16 @@ namespace Astroid {
 
       ustring tags_s;
 
+      vector<ustring> tags = m->tags;
+      for (ustring &tag : tags) {
+        tag = Glib::Markup::escape_text (tag);
+      }
+
 # ifndef DISABLE_PLUGINS
-      if (!thread_view->plugins->format_tags (m->tags, "#ffffff", false, tags_s)) {
+      if (!thread_view->plugins->format_tags (tags, "#ffffff", false, tags_s)) {
 #  endif
 
-        tags_s = VectorUtils::concat_tags_color (m->tags, false, 0, cv);
+        tags_s = VectorUtils::concat_tags_color (tags, false, 0, cv);
 
 # ifndef DISABLE_PLUGINS
       }
@@ -434,7 +439,7 @@ namespace Astroid {
 
       msg.set_tag_string (tags_s);
 
-      for (ustring &tag : m->tags) {
+      for (ustring &tag : tags) {
         msg.add_tags (tag);
       }
     }
@@ -464,15 +469,9 @@ namespace Astroid {
       if (static_cast<int>(bp.size()) > MAX_PREVIEW_LEN)
         bp = bp.substr(0, MAX_PREVIEW_LEN - 3) + "...";
 
-      while (true) {
-        size_t i = bp.find ("<br>");
-
-        if (i == ustring::npos) break;
-
-        bp.erase (i, 4);
-      }
-
-      msg.set_preview (Glib::Markup::escape_text (bp));
+      bp = UstringUtils::replace (bp, "<br>", "");
+      bp = UstringUtils::replace (bp, "\n", "");
+      msg.set_preview (bp);
     }
 
     if (astroid->config().get<std::string> ("thread_view.preferred_type") == "plain" &&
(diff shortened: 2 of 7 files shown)


ibuclaw:
I was curious to give this a whirl.  The iframe height is set too high a value.

image

Only after toggling between text/plain and text/html does the height correctly readjust itself.

jorsn:
Strange, for me it display fine.

Do you display plain or html by default?
Do you have a sample email for which this happens?

ibuclaw:
> Strange, for me it display fine.
> 

```
    "thread_view": {
        "open_html_part_external": "false",
        "preferred_type": "plain",
        "preferred_html_only": "true",
        "allow_remote_when_encrypted": "false",
        "open_external_link": "xdg-open",
        "default_save_directory": "~",
        "indent_messages": "false",
        "gravatar": {
            "enable": "true"
        },
        "mark_unread_delay": "0.5",
        "expand_flagged": "false"
    },
```

> Do you display plain or html by default? Do you have a sample email for which this happens?

i.e: All mail I get from GitHub notifications.

Ah!  I've just checked again, and it only affects messages that are collapsed when opening in the thread (i.e: `unread` tag removed).  Expanding them is what triggers the above screenshot.

(I'm also running with a variant of #745 as astroid no longer compiles on my system, so it might only be a GTK4 thing).

gauteh:
Hi. I don't have much capacity to test, if some of you want to help maintain I can add you.

jorsn:
Alright, I can now also reproduce it. Is there any way to inspect the html/css/js?
Debugging this is next to impossible without. Somehow it seems that no event besides `load` fires at all, for anything that happens to the page. In principle, resize events for all elements of class `email` should work in principle, since as far as I read, `display: none` sets the dimensions to zero.

ibuclaw:
> @ibuclaw I pushed a new commit that should fix the over-sizing. Does it work for you now?

Looks good to me.

gauteh:
The only one is that a sender could add javascript in the subject or in the email address. But I think that is possible to sanitize. As mentioned previously, I would be happy to give someone else access to merge.

jorsn:
> The only one is that a sender could add javascript in the subject or in the email address. But I think that is possible to sanitize.

Isn't this already sanitized by `Glib::Markup::escape_text(...)`? I checked every call to `webkit_dom_element_set_inner_html`, and I am pretty confident that every occurrence is sanitized in that way if it appears outside `.body_iframe`s in the end and contains data (in contrast to user-supplied config and program-generated things like error messages).

> As mentioned previously, I would be happy to give someone else access to merge.

I could take it, and help a bit with maintenance, especially when something breaks for me, but I can't tell you how much and for how long: I'm only willing to do it as long as I continue to use astroid, and I cannot give any guarantees on my time budget for this.
Further, I don't have any substantial C++, web or browser development experience, just general programming now and then.

gauteh:
Thanks, i appreciate that and am happy for what you can and have time to do.

gauteh:
I think sanitize should do the trick 👍

jorsn:
What is 'sanitize'? The JS method? How would you invoke it safely, here?

I think, *if* `escape_text()` is insufficient, the way to go is to set `innerText` insteas of `innerHTML`. For this, one has to generate the header table without the data first, and then populate with the data.

gauteh:
Innertext is safer yes. I was using sanitize and escape interchangeably.

jorsn:
I changed the setting of headers and preview to use innerText.
I think, this can be merged, but I don't seem to have write access to the repo, even if I am in the organization, now. At least, I cannot accept PRs.

gauteh:
Looks good. I tried adding you to this repo as well, please try again.

ibuclaw:
@jorsn thanks for tackling this!

ui/thread-view.html
```diff
@@ -57,9 +57,51 @@
...
+  <!--
+    sandbox="allow-same-origin" is required, otherwise the scrollHeight of the
+    embedded document cannot be accessed.
+    Do not add 'allow-scripts' to the sandbox attribute. This would allow the
+    document to execute scripts itself.
+  -->
+  <iframe class="body_iframe" sandbox="allow-same-origin" onload="handleLoadIframe(this);" srcdoc=""></iframe>
 </div>
 
+<script>
```
mcatanzaro, line 69: There's also a `WebKitSettings:enable-javascript-markup` setting you can use to remove any doubt; that allows the application to run JavaScript but blocks the document from doing so. That would block the use of `<script>` here, but you could probably run it manually using webkit_web_view_evaluate_javascript().

jorsn: This seems good for (faith in) security. But I did not see how I can reference `this` (the iframe) in cpp in a way such that I can pass it to the javascript function `handleLoadIframe`.

At least, I expect it to be a bit complicated.

Are your doubts strong, currently?

mcatanzaro: > But I did not see how I can reference this (the iframe) in cpp in a way such that I can pass it to the javascript function handleLoadIframe.

The best solution is the reverse: don't try to access it from C++. Move work that touches the DOM into JavaScript instead. The C++ DOM APIs are all deprecated and already gone in the webkitgtk-6.0 API version, so they'll be a problem for you in the future.

> Are your doubts strong, currently?

No. I _think_ the sandbox attribute is good enough because you're careful to encode everything so it can't break out. I think you're good so long as you use `Glib::Markup::escape_text` on all the untrusted input, and your comment says you do that.

[This guideline](https://cheatsheetseries.owasp.org/cheatsheets/DOM_based_XSS_Prevention_Cheat_Sheet.html#guideline-4-avoid-sending-untrusted-data-into-html-rendering-methods) suggests you don't use inner HTML at all, but if all untrusted input is properly encoded then I'm not sure why not.

jorsn: Thanks! Yes, in the long run, this schould be done, maybe one can just pass the emails as json to the javascript somehow, (anyway, it seems you can only pass data to js by injecting js strings, so it would be good to have *dead* json) and then use js to build the whole threadview. But this one is even more work.

PR #748, opened 2024-06-04
improve mailto uri handling and support webk2gtk 4.0 and 4.1

### LINKED ISSUE #608: astroid.desktop doesn't work with mailto: links
When setting astroid.desktop as the default handler for mailto: links, astroid opens, but without composing a new message. This is because astroid expects the email address to be passed to the --mailto flag, but this argument isn't present in the .desktop file. Adding it there breaks the "normal" way of opening astroid.

Possible fix: if only a single positional argument is given to astroid, treat it as a mailto: link and handle accordingly.

Related to #11.

### LINKED ISSUE #744: Stop using libsoup2.4 and webkit2gtk 4.0

jorsn:
fixes  #608, fixes #744
closes #745

I would be happy about comments on the mailto handling, which is described below.

#### mailto uri handling

@larskotthoff How do your problems trying to fix #608 relate to the approach taken here?

1. This introduces consistent mailto uri handling, supporting passing a body and all headers that can be given as agruments to EditMessage, namely, To, From, Cc, Bcc and Subject. Other headers are not dropped but prepended to the email body with a note.

2. Positional (i.e., non-option) commandline arguments are now interpreted as `--mailto` options.

3. Multiple `--mailto` options on the commandline are now supported, and are merged into one mailto argument before being passed to EditMessage.

4. A consequence is that libsoup is not used any more, enabling compatibility with webkit2gtk-4.1 while maintaining compatibility with gio (glib) 2.16, and thus building on older and current distros. Whereas supporting old distros is not necessary, it comes at very low cost: The only feature of newer gio that is used in #745 is the extraction of the path (To:) part and the query from the uri. But this is simply removing the scheme (`mailto:`) and splitting at `?`.

#### webkit2gtk 4.0 and 4.1

This is from https://github.com/astroidmail/astroid/commit/7ac88057441f432dbe44ea478e23dbfda70d32ac by @oreo639, but only the part that is necessary after https://github.com/astroidmail/astroid/commit/4c76a4cf6226752f5270197ba8747c7c7550e35e.

Version 4.1 is used by default. If this is unavailable, 4.0 is the fallback.


jorsn:
last force-push only fixed the commit message of https://github.com/astroidmail/astroid/pull/748/commits/c72b6fd93b6921d460d28f66620a130885137dd8.

jorsn:
@gauteh @oreo639 @ibuclaw If there are no comments comparing this to #745, I will merge this and close #745.

oreo639:
Guri was added in glib 2.66.2 not 2.16

jorsn:
I only use [g_uri_parse_scheme](https://docs.gtk.org/glib/type_func.Uri.parse_scheme.html) and [g_uri_unescape_string](https://docs.gtk.org/glib/type_func.Uri.unescape_string.html), which were added in version 2.16, according to the docs.

oreo639:
Ah, alright. Thanks.

PR #751, opened 2024-06-05
Fix help message for actions on multi-message-selection

ibuclaw:
This reverts commit 8f2a8aeefd634c3f861ba51f7ca5fe256c7350b1.

There is no other place in the code that calls `rev_multi->set_reveal_child(true)`, so the consequence of this change was that the `;` key no longer shows the revealer bar.

Before this PR:
image

After this PR:
image

### DIFF
--- a/src/main_window.cc
+++ b/src/main_window.cc
@@ -788,10 +788,8 @@ namespace Astroid {
     multi_waiting = true;
     multi_keybindings = kb;
 
-#if 0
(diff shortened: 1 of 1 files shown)


jorsn:
last force-push was only a change to the commit message

jorsn:
@zxramozx Could you explain what this means and what it has to do with this pr?

PR #752, opened 2024-06-05
Set margin_top on the revealer bars in addition to margin_bottom

ibuclaw:
1. Sets equal margins on both top and bottom, which (to me) has a better aesthetic look to it.
2. I noticed that there was a typo in the code which meant the bulk update ("multi") revealer never got a margin set.

---

Before this PR:

image

image

---

After this PR:

image

image

---

### DIFF
--- a/src/main_window.cc
+++ b/src/main_window.cc
@@ -145,6 +145,7 @@ namespace Astroid {
 
     rev_yes_no->set_margin_top (0);
     rh->set_margin_bottom (5);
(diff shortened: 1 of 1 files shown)


ibuclaw:
After a little toying around, I've convinced myself that a slightly smaller margin for the yes/no buttons looks better.

Updated "After" screenshot:

image

jorsn:
Thanks!

PR #753, opened 2024-06-05
Default initialize all pointers in Astroid class to avoid segfault in dtor

ibuclaw:
Under certain circumstances - both deterministic and non-deterministic - astroid can segfault on exit.

For example, when there's another window already running:
```
$ astroid & bg
[2024-06-05 16:18:20.526090] [0x0000706fccadcec0] [info]    welcome to astroid! - v0.16-20-g41808e88
...
[16:18:20] [0x0000706fccadcec0] [M] [debug] log: stdout: yes

$ astroid 
[2024-06-05 16:20:26.487138] [0x00007f11b1bbaec0] [info]    welcome to astroid! - v0.16-20-g41808e88
Segmentation fault (core dumped)
```

In all cases the cause is at least one of these pointers pointing at garbage due to being uninitialised.

### DIFF
--- a/src/astroid.hh
+++ b/src/astroid.hh
@@ -63,17 +63,17 @@ namespace Astroid {
       ustring user_agent;
 
       /* accounts */
-      AccountManager * accounts;
+      AccountManager * accounts = NULL;
 
       /* actions */
-      ActionManager * actions;
+      ActionManager * actions = NULL;
 
 # ifndef DISABLE_PLUGINS
-      PluginManager * plugin_manager;
+      PluginManager * plugin_manager = NULL;
 # endif
 
       /* poll */
-      Poll * poll;
+      Poll * poll = NULL;
 
       MainWindow * open_new_window (bool open_defaults = true);
 
@@ -81,7 +81,7 @@ namespace Astroid {
       GdkAtom clipboard_target = GDK_SELECTION_CLIPBOARD;
 
     protected:
-      Config * m_config;
+      Config * m_config = NULL;
 
     private:
       void on_activate () override;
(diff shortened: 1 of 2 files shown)


ibuclaw:
Added fix for the uncaught exception reported in issue 727.

It turns out that libnotmuch 5.4 has a new API for reading configuration values so there's no need for Astroid to roll its own anymore - alas Debian bullseye has libnotmuch 5.3, otherwise I would have rewritten all this code.

gauteh:
Hi,

 I would recommend not to accommodate old debians. Don't have time to
support so many different versions.

tor. 6. juni 2024, 15:55 skrev Iain Buclaw ***@***.***>:

ibuclaw:
> Hi, I would recommend not to accommodate old debians. Don't have time to support so many different versions.

Sure, have nothing to disagree about that, my thinking is that fixing the issue and setting a new minimum version requirement for libnotmuch (5.4 - or notmuch 0.32) are two separate things.

There is a new function in [libnotmuch 5.4](https://github.com/notmuch/notmuch/blob/e8e2d5247e44d9b402cbcc491b859cd9398f045d/lib/notmuch.h#L468-L499) - however migrating to using it would mean a rewrite away from using ptree to using libnotmuch API directly for reading the notmuch ini file (i.e: [to read a boolean value](https://github.com/notmuch/notmuch/blob/e8e2d5247e44d9b402cbcc491b859cd9398f045d/lib/notmuch.h#L2771-L2789))

Adding in a new catch handler in the meantime is a trivial fix.

jorsn:
> my thinking is that fixing the issue and setting a new minimum version requirement for libnotmuch (5.4 - or notmuch 0.32) are two separate things.
-- @ibuclaw 

I agree, since my first priority is to release a version that works. Then, one can still modernize and clean up.

@gauteh Any objections against merging this ?

gauteh:
Sounds good. Note that there has been a bunch of issues in lieer regarding the new config system, where some values are in the db, some in the config file. Doesn't appear to work perfectly always. But probably not something astroid should worry about.

PR #754, opened 2024-06-09
Stop using deprecated Gtk::HBox

ibuclaw:
GtkHBox has been deprecated since GTK 3.2.  No further changes are needed, since the default value of the GtkOrientable:orientation property is GTK_ORIENTATION_HORIZONTAL.

(See: https://docs.gtk.org/gtk3/class.HBox.html)


jorsn:
Thanks!

PR #755, opened 2024-06-09
Call set_size_request on pane children

### LINKED ISSUE #547: Astroid pane view doesn't work anymore
When trying to open pane view in astroid version 8cd76eeb9b, I only get a small supplementary blank space (~1mm height) before the first thread, and the the thread tab freeze.

ibuclaw:
Otherwise the second child gets no width, and appears to be hidden when a thread is opened in pane view.

The number `300` is pretty arbitrary, it's taken as 1/4 of the default window width ([set here](https://github.com/astroidmail/astroid/blob/c1e5cdbd662e2bcfef2fe5dc72dbc444a692a0e8/src/main_window.cc#L105)).

Fixes #547

### DIFF
--- a/src/modes/paned_mode.cc
+++ b/src/modes/paned_mode.cc
@@ -55,7 +55,7 @@ namespace Astroid {
 
     Gtk::EventBox * fb = Gtk::manage (new Gtk::EventBox ());
     fb->add (*w);
-    fb->set_size_request (300);
+    fb->set_size_request (50);
 
     if (p == 0) {
       fp1 = fb;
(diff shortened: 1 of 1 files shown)


jorsn:
Thanks, this works, and by default sets the pane size to half the window width.
300 seems pretty high to me, though, as it is the minimal width. Wouldn't 50 be enough?

Strangely enough, with only 20, the default width was smaller for me, maybe around the 300.

ibuclaw:
I can't remember the effect of setting it to 50. Did initially toy with various sizes, I noticed that when too small, it didn't split the window evenly into two, rather split 2/3 and 1/3. Value I ended up with I think is a quarter the default width. On my screen any smaller and threads become mostly unreadable, but with a different font that would obviously change.

